PR #126, opened 2014-05-11
Handlebars recipe

callumlocke:


codeofsumit:
I get the following error when following this recipe:

```
ncaught Error: Template was precompiled with an older version of Handlebars than the current runtime. Please update your precompiler to a newer version (>= 4.0.0) or downgrade your runtime to an older version (>= 2.0.0-beta.1).
```

I'm not (don't want to) use any precompilation and I'm note sure where that comes from actually.
Any ideas what might have changed that isn't reflected in the recipe yet?

silvenon:
@codeofsumit can you open an issue?

codeofsumit:
sure thing, #480

PR #266, opened 2015-02-03
Added woff2 to the list of font files to seek

doublemarked:
The Bootstrap 3 package references a woff2 file for glyphicons, but the gulpfile was not including this extension in its font bundle.


silvenon:
Good catch, thanks! :+1:

PR #268, opened 2015-02-05
Remove `minimal-ui` viewport property

arthurvr:
The `minimal-ui` viewport property was introduced in iOS 7.1, but is no longer supported in iOS 8.

Ref https://developer.apple.com/library/prerelease/ios/releasenotes/General/RN-iOSSDK-8.0/#//apple_ref/doc/uid/TP40014223-CH1-SW83

### DIFF
--- a/app/templates/index.html
+++ b/app/templates/index.html
@@ -3,7 +3,7 @@
   <head>
     <meta charset="utf-8">
     <meta name="description" content="">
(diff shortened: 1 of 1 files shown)


arthurvr:
Again, no idea why tests are failing for this one.

silvenon:
I broke the tests a while ago, so it's not your fault. I'll fix those ASAP.

PR #272, opened 2015-02-09
Revert to Ruby Sass and add sourcemaps

### LINKED ISSUE #256: Sass sourcemap
The Sass sourcemap is being generated under .tmp. But, why is it not being exposed to Web Inspector? 
I must be missing something because the same "issue" occurs using the grunt version of generator webapp.

silvenon:
Sourcemaps will be generated without Sass as well, for PostCSS.

Fixes #256.

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,19 +11,13 @@ Please see our [gulpfile.js](app/templates/gulpfile.js) for up to date informati
 
 * CSS Autoprefixing
 * Built-in preview server with BrowserSync
-* Automagically compile Sass with [libsass](http://libsass.org)
+* Automagically compile Sass
+* Map compiled CSS to source stylesheets with source maps
 * Automagically lint your scripts
 * Awesome image optimization
 * Automagically wire-up dependencies installed with [Bower](http://bower.io)
 
 *For more information on what this generator can do for you, take a look at the [gulp plugins](app/templates/_package.json) used in our `package.json`.*
-
-
-## libsass
-
-Keep in mind that libsass is feature-wise not fully compatible with Ruby Sass. Check out [this](http://sass-compatibility.github.io) curated list of incompatibilities to find out which features are missing.
-
-If your favorite feature is missing and you really need Ruby Sass, you can always switch to [gulp-ruby-sass](https://github.com/sindresorhus/gulp-ruby-sass) and update the `styles` task in `gulpfile.js` accordingly.
 
 
 ## Getting Started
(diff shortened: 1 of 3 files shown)


silvenon:
Oh, I forgot to add. I'm reverting to Ruby Sass because I don't know when [libsass 3.2](https://github.com/sass/libsass/issues?q=is%3Aopen+is%3Aissue+milestone%3A3.2) will be released and I want something stable.

sindresorhus:
:-1:

silvenon:
Why? I really like it.

silvenon:
Not sure why tests fail, they ran successfully locally.

sindresorhus:
I don't see why we would want to revert back to ruby sass again. Source maps are useful, but not that important. I'd rather have speed.

silvenon:
I'd rather have merged `@media` queries. While nested `@media` queries are valid CSS3, they appear to not work in IE11 (and maybe in other browsers which otherwise support `@media` queries). http://codepen.io/silvenon/pen/JoOxWb

I reverted simply because I wouldn't use this myself in the current state and I want to make a stable release. I'm planning to add libsass back as soon as I'm happy with it.

sindresorhus:
It's up to you. Feel free to merge if you think it's the best.

I just don't care about IE.

silvenon:
I only care about IE10+ and I think that's as far as we should assume for our users.

Ok, I'll fix the tests and merge.

addyosmani:
I'm +1 for speed over source maps on this one. I can only imagine the majority of devs would feel the same, but perhaps consider asking folks if they'd prefer one over the other before landing? The differences during watch are particularly large.

mattycraig:
+1 for speed for sure.

austinpray:
FYI: node-sass is currently broken on 0.12. Also for the past couple months it intermittently fails to build on linux. 

The instability is a drag, but I would have a hard time going back to ruby sass. I'm hoping [I can spend some time this week](https://github.com/sass/node-sass/issues/602) to help the node-sass guys get 2.0 out the door.

silvenon:
Speed > support for any IE at all? About 17% of web traffic?

austinpray:
Not supporting latest IE is some silly stuff.

thasmo:
+1 libsass

arthurvr:
I must say I'm against dropping libsass too.

mu7amad:
+1 libsass

dbashford:
Largely unfamiliar with Yeoman, so sorry if I'm out in left field, but isn't this something you can ask the user during scaffold?  Probably a goodly amount of rework to make that happen, but if folks are evenly split, if its possible...  a worthwhile effort?

silvenon:
@dbashford I think libsass is too close to being stable for this extra work to be worth it, but I can do it if more people would like this feature.

silvenon:
Ok, everyone since to be against temporarily dropping libsass, so I'm closing this in favor of #274, because I mainly wanted source maps anyway.

Issue #159, opened 2014-06-21
fonts task not ending properly?


xmlking:
Unlike other tasks `fonts` doesn't seems ending properly. Why `fonts` task not printing `Finished` statement?
This is causing my build task fail. I noticed this behavior after updating my package.json to latest versions ( gulp, gulp-filter, gulp-bower-files ... )

This fails...

``` js
gulp.task('build', ['jshint', 'html', 'images', 'fonts', 'extras'], function () {
    return gulp.src('dist/**/*').pipe($.size({title: 'build', gzip: true}));
});
```

but this works...

``` js
gulp.task('build', ['jshint', 'html', 'images', 'extras'], function () {
    return gulp.src('dist/**/*').pipe($.size({title: 'build', gzip: true}));
});
```

`fonts` task output: (no `Finished` statement in the output)

```
C:\Work\demo>gulp fonts
[16:14:22] Using gulpfile C:\Work\demo\gulpfile.js
[16:14:22] Starting 'fonts'...
```

`styles` task output:

```
C:\Work\demo>gulp styles   
[16:15:26] Using gulpfile C:\Work\demo\gulpfile.js
[16:15:26] Starting 'styles'...
[16:15:29] Finished 'styles' after 2.93 s
```

Environment 

```
  "gulp": "^3.8.1",
  "gulp-filter": "^0.5.0",
  "gulp-flatten": "^0.0.2", 
  "gulp-bower-files": "^0.2.5",
```

``` js
gulp.task('fonts', function () {
  var streamqueue = require('streamqueue');
  return streamqueue({objectMode: true},
      $.bowerFiles(),
      gulp.src('app/fonts/**/*')
    )
    .pipe($.filter('**/*.{eot,svg,ttf,woff}'))
    .pipe($.flatten())
    .pipe(gulp.dest('dist/fonts'));
});
```

Error:

```
undefined:1672
      for (var $__19 = iterable[Symbol.iterator](),
                                                ^
TypeError: Property 'Symbol()' of object  is not a function
    at new Set (eval at <anonymous> (C:\Work\demo\node_modules\gulp-traceur\node_modules\traceur\src\node\traceur.js:24:17), <anonymous>:1672:49)
    at uniqSet (C:\Work\demo\node_modules\gulp-filter\node_modules\multimatch\node_modules\array-union\node_modules\array-uniq\index.js:18:3)
[...]

PR #277, opened 2015-02-12
Server fixes

silvenon:
1. I've encountered #159 myself recently. Updating main-bower-files and getting rid of gulp-filter solved it. I also removed gulp-flatten as it's unnecessary.
2. Now changes to `app/fonts/**/*` will trigger the `fonts` task, which fixes the issue where the user would not see updates made in `app/fonts` because the fonts in `.tmp/fonts` (with the same name) have precedence.
3. There was redundancy around BrowserSync reloading, there's no need to refresh the browser when a task is triggered, only when actual files change, it's more semantic.
4. Added CSS live reloading.

If you agree about these changes, I'll make necessary recipe updates.

### DIFF
--- a/app/templates/gulpfile.js
+++ b/app/templates/gulpfile.js
@@ -21,7 +21,8 @@ gulp.task('styles', function () {<% if (includeSass) { %>
       require('autoprefixer-core')({browsers: ['last 1 version']})
     ]))
     .pipe($.sourcemaps.write())
-    .pipe(gulp.dest('.tmp/styles'));
+    .pipe(gulp.dest('.tmp/styles'))
+    .pipe(reload({stream: true}));
 });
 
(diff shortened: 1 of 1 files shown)


sindresorhus:
Looks good :)

timbog80:
why hasnt this been fixed in yeoman yet?

silvenon:
What does that mean? You mean released?

michelmany:
I still having a issue with fonts task. :(

silvenon:
@michelmany which issues? What version of the generator are you using?

michelmany:
sorry @silvenon silvenon ... its ok now... the directory in the .bowerrc was incorrect. thanks!

silvenon:
:+1:

PR #293, opened 2015-03-22
Added asset revision.

hivre:
See the discussion at #255 this is the napkin-solution. 
I have made the least changes to the current setup of the gulp file.
- adds two new plugins: rev-all and rev-napkin
  - Rev-all takes care of adding revisions numbers to assets.
  - Rev-Napkin removes the (duplicate) un-revved files from the stream
- moved HTML minification to the build task, as it interferes with the revision management
- added default ignores for revisons: all html files, favicon, touch-icons and robots.txt

### DIFF
--- a/app/templates/gulpfile.js
+++ b/app/templates/gulpfile.js
@@ -42,7 +42,6 @@ gulp.task('html', ['styles'], function () {
     .pipe($.if('*.css', $.csso()))
     .pipe(assets.restore())
     .pipe($.useref())
-    .pipe($.if('*.html', $.minifyHtml({conditionals: true, loose: true})))
     .pipe(gulp.dest('dist'));
 });
 
@@ -131,7 +130,12 @@ gulp.task('wiredep', function () {
 });
 
 gulp.task('build', ['jshint', 'html', 'images', 'fonts', 'extras'], function () {
-  return gulp.src('dist/**/*').pipe($.size({title: 'build', gzip: true}));
+  return gulp.src('dist/**/*', {base: 'dist'})
+    .pipe($.revAll({ignore: ['favicon.ico', 'apple-touch-icon.png', 'robots.txt', '.html']}))
+    .pipe($.if('*.html', $.minifyHtml({conditionals: true, loose: true})))
+    .pipe($.revNapkin())
+    .pipe(gulp.dest('dist'))
+    .pipe($.size({title: 'build', gzip: true}));
 });
 
 gulp.task('default', ['clean'], function () {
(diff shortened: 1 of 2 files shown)


silvenon:
It seems to work correctly and gives a nice colorful output.

screenshot 2015-04-10 15 53 39

Two questions:

1. I noticed there's a new version of gulp-rev-all with some breaking changes, would you care to update to the latest version?
2. The output seems a bit too verbose. Perhaps silent is best? I think the new option is called `debug`, so perhaps set that to `false`.

sindresorhus:
@silvenon Not sure what more to say. I've already said I don't like this complication and I think it will cause a lot of support requests, but totally up to you. You're also not using `gulp-rev`, but `gulp-rev-all`.

silvenon:
Ok, I take full responsibility if it gets out of hand :smiley:

I'd like to continue with this then, so @hivre, when you have time, please address my 2 questions.

app/templates/gulpfile.js
```diff
@@ -131,7 +130,12 @@ gulp.task('wiredep', function () {
 });
 
 gulp.task('build', ['jshint', 'html', 'images', 'fonts', 'extras'], function () {
-  return gulp.src('dist/**/*').pipe($.size({title: 'build', gzip: true}));
+  return gulp.src('dist/**/*', {base: 'dist'})
+    .pipe($.revAll({ignore: ['favicon.ico', 'apple-touch-icon.png', 'robots.txt', '.html']}))
```
austinpray: https://github.com/smysnk/gulp-rev-all/blob/591d18d4247f287b0821896f8a266b6f7128963c/README.md#upgrading-to-v08x

>New Feature: Ignoring files has changed, ignore option has been removed and has been replaced with dontGlobal, dontRenameFile, dontUpdateReference, dontSearchFile which allows for more control and less ambiguity on what is being ignored

arthurvr:
> I'd like to continue with this then, so @hivre, when you have time, please address my 2 questions.

ping @hivre :)

arthurvr:
ping @hivre - also needs a rebase :)

silvenon:
We'll have to continue this in another PR, one of these days :smiley:

hivre:
I'm sorry for the lack of communications. Startup collapsed, new job busy which didn't have any programming. 

If it is still needed, i'm happy to finish this. I noticed a new version of the revving plugins that solve some issues we had before.

silvenon:
Sure, if you're up for finishing, I'd be happy to merge :wink:

hivre:
Great.
I'll pick this up this weekend.

silvenon:
No pressure.

hivre:
see #360 for a new attempt.

from an earlier comment:

> 1. I noticed there's a new version of gulp-rev-all with some breaking changes, would you care to update to the latest version?
 _done_
> 2. The output seems a bit too verbose. Perhaps silent is best? I think the new option is called debug, so perhaps set that to false.
_done_ (set debug: true in the options list to get it back)

Close this one?

silvenon:
:+1:

PR #300, opened 2015-04-06
Added clearing of gulp-cache to clean task

szebrowski:
I've changed some image's names and even after gulp clean I got old filenames in dist/ dir. Adding clearing of gulp-cache fixed issue.

### DIFF
--- a/app/templates/gulpfile.js
+++ b/app/templates/gulpfile.js
@@ -75,7 +75,10 @@ gulp.task('extras', function () {
   }).pipe(gulp.dest('dist'));
 });
 
-gulp.task('clean', require('del').bind(null, ['.tmp', 'dist']));
+gulp.task('clean', function(done){
+  require('del')(['.tmp', 'dist']);
+  return $.cache.clearAll(done);
+});
 
 gulp.task('serve', ['styles', 'fonts'], function () {
   browserSync({


app/templates/gulpfile.js
```diff
@@ -75,7 +75,10 @@ gulp.task('extras', function () {
   }).pipe(gulp.dest('dist'));
 });
 
-gulp.task('clean', require('del').bind(null, ['.tmp', 'dist']));
+gulp.task('clean', function(done){
+  require('del')(['.tmp', 'dist']);
+  return $.cache.clearAll(done);
+});
```
sindresorhus: `del` is async, so this is incorrect as it will not wait for it to finish.

app/templates/gulpfile.js
```diff
@@ -75,7 +75,10 @@ gulp.task('extras', function () {
   }).pipe(gulp.dest('dist'));
 });
 
-gulp.task('clean', require('del').bind(null, ['.tmp', 'dist']));
+gulp.task('clean', function(done){
```
arthurvr: code style

```js
function (done) {
```

silvenon:
Then there's not much point in using gulp-cache at all, is there? Because the `images` task is only run on build and every time the cache will be cleared.

sebastianzebrowski:
Yeah, so I purpose to switch to https://www.npmjs.com/package/gulp-newer

silvenon:
I don't see how gulp-newer is going to help here, we don't want to optimize only newer images, because old images won't end up in the build then.

I'm closing this because clearing the cache every time would render gulp-cache useless. This should be done manually, in rare circumstances when something went wrong.

PR #301, opened 2015-04-07
Update to match current generated files

GeneralStrike:
Removed ref to 'reload' to match current auto-generated gulpfile.js


silvenon:
Thanks :smiley:

PR #305, opened 2015-04-20
gulp task "styles" use src: *.scss/css instead of main.scss/css only

DanweDE:
Don't see why this shouldn't be done. Had a little struggle adding another .scss and getting it into the built. Adding it in the built-block comment in the index.html had no effect because of the way the gulp task was designed. Now it works when having the following in index.html for example:

```
<!-- build:css styles/main.css -->
<link rel="stylesheet" href="styles/main.css">
<link rel="stylesheet" href="styles/something-else.css">
<!-- endbuild -->
```

### DIFF
--- a/app/templates/gulpfile.js
+++ b/app/templates/gulpfile.js
@@ -7,7 +7,7 @@ var browserSync = require('browser-sync');
 var reload = browserSync.reload;
 
 gulp.task('styles', function () {<% if (includeSass) { %>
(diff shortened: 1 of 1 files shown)


silvenon:
I agree, by having compiling only `main.scss` we're losing flexibility for no reason, and without Sass it makes no sense at all, because you'll most probably want more stylesheets.

Thanks for the PR :+1:

PR #312, opened 2015-05-07
Include wiredep as a dependency to serve

aweary:
I was running into an issue where I'd stop the Gulp task, install something with Bower, and then start Gulp back up again. Every time I did this, wiredep wouldn't run and my new dependency would be ignored.

I realize this is because I'm not touching the bower.json file _while_ the Gulp process is running. I usually use one Terminal window at a time, so if we just included wiredep as a dependency to serve, we could be sure that any new dependencies added since the last run would be included.

It may be wise to include it in the build task as well.

### DIFF
--- a/app/templates/gulpfile.js
+++ b/app/templates/gulpfile.js
@@ -147,7 +147,7 @@ gulp.task('wiredep', function () {
 <% } %>
   gulp.src('app/*.html')
     .pipe(wiredep({<% if (includeSass && includeBootstrap) { %>
-      exclude: ['bootstrap-sass-official'],<% } %>
+      exclude: ['bootstrap-sass'],<% } %>
       ignorePath: /^(\.\.\/)*\.\./
     }))
     .pipe(gulp.dest('app'));


silvenon:
I agree that we need some kind of a solution for this, because it's a gotcha. But I think this way, where `styles` and `wiredep` are ran simultaneously, `wiredep` will inject Sass dependencies, but because `styles` already started (or even finished) they won't be compiled that first time.

Aweary:
That's a good point. I suppose some other options would be to add `wiredep` as a dependency to `styles` or maybe use something like `run-sequence` to ensure that the tasks are ran in the correct order. 

The former doesn't seem idiomatic though, and will needlessly invoke `wiredep` a lot. The later could be accomplished using an alias task of some sort like:

 ```js
gulp.task('compile', function(){
    $.runSequence(['wiredep', 'styles', 'fonts']).
  });
```

```js
gulp.task('serve', ['compile'], function(){ ... });
```

But that involves adding another dependency to the Gulp.

silvenon:
Yeah, I'm reluctant to further complicate `gulpfile.js`, as this is not a fatal issue, you can just run `gulp wiredep` yourself if you installed some deps while gulp wasn't watching.

I suggest that in future you create separate branches for your PRs, because now your new commits on master entered this PR :smiley:

If that was intentional, I'd like to keep things separate, so please pick one change set (e.g. bootstrap-sass), move others to a separate branch, and we can rename the PR :wink:

silvenon:
Oh, you already did.

Aweary:
Yeah, sorry about that, I still have my Github training wheels on :smiley:

silvenon:
We have it documented [here], in our Getting Started guide. Do you think it's hard to find and we could improve it somehow?

Aweary:
Nope, I'm just bad at finding things! Thank you.

PR #313, opened 2015-05-08
Update gulp-sass

febLey:
A new version of gulp-sass including node-sass v3.0.0 and libsass v3.2.0 has been released.


silvenon:
Oh yeah!

PR #315, opened 2015-05-08
Updated outputStyle to expanded

antleblanc:
According to the new Libsass release.

You can find more information by following this url :
https://github.com/sass/libsass/releases/tag/3.2.0


sindresorhus:
The `node-sass` readme still says that it's broken, so would be nice if you could update that too ;)

antleblanc:
Thanks for pointing it out to me, it's done ;)

PR #332, opened 2015-06-09
Gulp Inject compiled coffeescript files into html

devil1991:


silvenon:
This would introduce inconsistencies with our recipes (and this generator in general), because we don't use gulp-inject anywhere else. Also, what if you need to control the order in which files are injected? The order mostly doesn't matter with vendor libraries, but it sometimes matters with your own files.

silvenon:
I'll reopen if convincing arguments are provided.

PR #339, opened 2015-06-17
Fixes lint:test and serve:test with the default files

tmorton:
I ran into problems running tests with a newly-generated app.  This PR fixes those issues.
- The `lint:test` task needs some options to pass the default mocha test file.
- The `serve:test` task needs a route to the bower components
- The `serve:test` task should open a browser like the other `serve` tasks.

### DIFF
--- a/app/templates/gulpfile.babel.js
+++ b/app/templates/gulpfile.babel.js
@@ -35,13 +35,13 @@ function lint(files, options) {
   };
 }
 const testLintOptions = {
-  "env": {
-    "mocha": true
+  env: {
+    mocha: true
   },
-  "globals": {
-    "assert": false,
-    "expect": false,
-    "should": false
+  globals: {
+    assert: false,
+    expect: false,
+    should: false
   }
 };
 


app/templates/gulpfile.babel.js
```diff
@@ -25,18 +25,28 @@ gulp.task('styles', () => {<% if (includeSass) { %>
...
   };
 }
+const testLintOptions = {
+  "env": {
+    "mocha": true
+  },
+  "globals": {
+    "assert": false,
+    "expect": false,
+    "should": false
```
kevva: Remove `"` from the keys.

app/templates/gulpfile.babel.js
```diff
@@ -123,11 +133,13 @@ gulp.task('serve:dist', () => {
 gulp.task('serve:test', () => {
   browserSync({
     notify: false,
-    open: false,
```
kevva: Why?

tmorton:
I'm not sure why Travis is failing.  It's only failing the Node 0.10 check, which passes on my machine.

silvenon:
Maybe better to have those options in the `package.json` where the rest of them are? What do you think @kevva?

silvenon:
Btw, I restarted the test and now it passes. It sometimes times out, we should probably increase the timeout.

tmorton:
@silvenon If they're in `package.json` won't they apply to the regular `lint` task as well?  That seems undesirable.

silvenon:
You're right, it's best not to mix them.

silvenon:
LGTM, thanks! :+1:

PR #359, opened 2015-07-27
avoid main-bower-files spitting error when there is no bower deps

stevemao:
When there is no bower deps it will error when running `gulp`. You have to create an empty bower_components by hand. This fix avoids it.

Ref: https://github.com/ck86/main-bower-files/issues/103


stevemao:
travis fails due to timeout. maybe increase the timeout? shouldn't have anything to do with this one.

silvenon:
Thanks!

PR #360, opened 2015-08-03
Added improved revision functionality.

hivre:
**Quick Overview**
1. new `constant` revAll to get all the options possible for revving
2. extended the build task to perform revving
3. all files ready for revving are moved to the .tmp directory
4. rev-all searches in .tmp for files, revvs them and moves them to dist folder

**Elaboration**

**1. new `constant`**
The new version of rev-all (> 0.8) needs an instantiation. Furthermore, many more options have been added for customization. 
I thought of making a new revving task, which could be re-used (e.g for deployment), but per scenario a user probably wants different options, which means a new instantiation, and thus a new task.
A constant is the neatest/most compact option from my perspective.

**2. / 3. New build task & .tmp**
Obviously the build task needed to get extended, as we want to perform all other tasks before revving.
The current setup, contrary to the previous, follows the much discussed method of prepping everything into `.tmp`, and then getting the final files stored into `dist`. (see the other PR and issue for the discussion).
- Each task that would store files into the `dist` folder, now does that solely in `.tmp`.
- After all build-tasks are completed, `rev-all` goes over the files in `.tmp`, revvs them accordingly, and stores them in `dist`.
- two major advantages:  1. Only final-ready-to-deploy files are in the `dist` folder. 2. no need for `rev-napkin` and the like to remove un-revved and pre-revved (thus temporary) files. 

**Side note**
I normally test my changes on my own website, however, that is still using v.0.3 and 
a. I don't know how to elgantly upgrade the generator inside that project
b. it's 2.30am so not in the mood to figure that out right now.

I tested this using the standard yeoman/generator webpages that come along with the generator. I don't expect anything weird to happen when I use this on a larger project (famous last words, lol)

Looking at the amount of changes since the first RP I did, I would propose to close that one. Consensus was that this (all to `.tmp`, then to `dist`) was the best method anyway...

### DIFF
--- a/app/templates/_package.json
+++ b/app/templates/_package.json
@@ -15,7 +15,8 @@
     "gulp-imagemin": "^2.2.1",
     "gulp-load-plugins": "^0.10.0",
     "gulp-minify-css": "^1.1.1",
-    "gulp-minify-html": "^1.0.0",<% if (includeSass) { %>
+    "gulp-minify-html": "^1.0.0",
+    "gulp-rev-all": "^0.8.0",<% if (includeSass) { %>
     "gulp-plumber": "^1.0.1",
     "gulp-sass": "^2.0.0",<% } %>
     "gulp-size": "^1.2.1",
--- a/app/templates/gulpfile.babel.js
+++ b/app/templates/gulpfile.babel.js
@@ -7,6 +7,7 @@ import {stream as wiredep} from 'wiredep';
 
 const $ = gulpLoadPlugins();
 const reload = browserSync.reload;
+const revAll = new $.revAll({dontRenameFile: ['favicon.ico', 'apple-touch-icon.png', 'robots.txt', '.html']});
 
 gulp.task('styles', () => {<% if (includeSass) { %>
   return gulp.src('app/styles/*.scss')
@@ -57,7 +58,7 @@ gulp.task('html', ['styles'], () => {
     .pipe(assets.restore())
     .pipe($.useref())
     .pipe($.if('*.html', $.minifyHtml({conditionals: true, loose: true})))
-    .pipe(gulp.dest('dist'));
+    .pipe(gulp.dest('.tmp'));
 });
 
 gulp.task('images', () => {
@@ -73,14 +74,13 @@ gulp.task('images', () => {
       console.log(err);
       this.end();
     })))
-    .pipe(gulp.dest('dist/images'));
+    .pipe(gulp.dest('.tmp/images'));
 });
 
 gulp.task('fonts', () => {
   return gulp.src(require('main-bower-files')('**/*.{eot,svg,ttf,woff,woff2}', function (err) {})
     .concat('app/fonts/**/*'))
     .pipe(gulp.dest('.tmp/fonts'))
-    .pipe(gulp.dest('dist/fonts'));
 });
 
 gulp.task('extras', () => {
@@ -89,7 +89,7 @@ gulp.task('extras', () => {
     '!app/*.html'
   ], {
     dot: true
-  }).pipe(gulp.dest('dist'));
+  }).pipe(gulp.dest('.tmp'));
 });
 
 gulp.task('clean', del.bind(null, ['.tmp', 'dist']));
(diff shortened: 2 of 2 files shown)


alferov:
I'm sorry, but `gulp-rev-all` is [blacklisted](https://github.com/gulpjs/plugins/blob/master/src/blackList.json#L70).

> "gulp-rev-all": "duplicate of gulp-rev",

[gulp-rev](https://github.com/sindresorhus/gulp-rev) should be able to get the job done.

app/templates/_package.json
```diff
@@ -15,7 +15,8 @@
     "gulp-imagemin": "^2.2.1",
     "gulp-load-plugins": "^0.10.0",
     "gulp-minify-css": "^1.1.1",
-    "gulp-minify-html": "^1.0.0",<% if (includeSass) { %>
+    "gulp-minify-html": "^1.0.0",
+    "gulp-rev-all": "^0.8.0",<% if (includeSass) { %>
```
alferov: I'm sorry, but `gulp-rev-all` is [blacklisted](https://github.com/gulpjs/plugins/blob/master/src/blackList.json#L70).

> "gulp-rev-all": "duplicate of gulp-rev",

[gulp-rev](https://github.com/sindresorhus/gulp-rev) should be able to get the job done.

hivre: Ok, I can work on that, but it'll require about 3 plugins to do what rev-all does. 
Interesting strategy to blacklist forks...

hivre:
I looked into using `gulp-rev`.
For it to replace `rev-all`, you need to use `gulp-rev`, `rev-replace`, `rev-css-url`. (I'm assuming the `.tmp` setup as in this PR works a expected)
It adds complexity but also flexibility. 

The problem I'm at, after 45mins of Googling, is how to make exceptions for e.g. favicon or index.html. I can't find it, unless you use a `filter()` function.

hivre:
@silvenon @alferov is the "no blacklisted modules" a hard line? 
Because with `gulp-rev` et al. it's going to be a lot more complicated...

app/templates/_package.json
silvenon: Asset revisioning is pretty complicated to set up with gulp, so I'm sad that they blacklisted it, but I guess composite plugins aren't best practice.

silvenon:
I think we should try to set it up with gulp-rev, let's see how we can make it as simple as possible.

silvenon:
Btw, thanks so much for helping us with this one.

sondr3:
What I did, which works very well is just dropping everything into the `styles` task in my gulpfile and then use `gulp-inject` to inject the CSS into the header. My task is probably quite a bit overkill for this (and I didn't bother changing the paths etc), but what it does is generate a `style.scss` file, rename it to `style.min.css` and then rev it so it gets renamed to something like `style-17ga7g6.min.css`, after which you create another gzipped version as well.

```js
gulp.task('styles', () =>
  gulp.src('src/assets/scss/style.scss')
    .pipe($.sourcemaps.init())
    .pipe($.sass({
      precision: 10
    }).on('error', $.sass.logError))
    .pipe($.postcss([
      autoprefixer({browsers: 'last 1 version'})
    ]))
    .pipe($.rename('style.min.css'))
    .pipe($.if('*.css', $.minifyCss()))
    .pipe($.size({
      title: 'minified styles',
      showFiles: true
    }))
    .pipe($.rev())
    .pipe(gulp.dest('.tmp/assets/stylesheets'))
    .pipe($.sourcemaps.write('.'))
    .pipe(gulp.dest('.tmp/assets/stylesheets'))
    .pipe($.if('*.css', $.gzip({append: true})))
    .pipe($.size({
      title: 'gzipped styles',
      gzip: true,
      showFiles: true
    }))
    .pipe(gulp.dest('.tmp/assets/stylesheets'))
);
```

Then in your `index.html` you create

```html
  <!-- inject:css -->
  <!-- endinject -->
```

in the head area which will be automatically populated with a task like this:

```js
gulp.task('inject', () =>
  gulp.src('src/_includes/head.html')
    .pipe($.inject(gulp.src('.tmp/assets/stylesheets/*.css',
                            {read: false}), {ignorePath: '.tmp'}))
    .pipe(gulp.dest('src/_includes'))
);
```

and you'll get

```html
  <!-- inject:css -->
  <link rel="stylesheet" href="/assets/stylesheets/style-17ga7g6.min.css">
  <!-- endinject -->
```

which will include a link to it's sourcemap and have a gzipped version alongside as well. It works very well for me, but obviously it's quite different from this one. I can make a PR if that makes things clearer.

silvenon:
Thanks for sharing your workflow! I guess it will need to be a bit of an overkill :confused:

sondr3:
Yeah, getting revving to work has been a major PITA for me and this was the best solution I could figure out.

sindresorhus:
Closing for lack of activity.

PR #386, opened 2015-09-22
Documentation update

alferov:
Hi, I've noticed that only Jade recipe was updated.
Other recipes deserved updates too! :smile:


silvenon:
Wow, thanks :+1:

PR #392, opened 2015-09-28
Recipe: Setting up React Router in addition to React and JSX

ghost:


evanheisler:
Would it be possible to see an example of how to use React Router in the project? The recipe works great and I have it installed, but am not sure how to utilize it as the integration seems quite different than the [React Router documentation](https://github.com/rackt/react-router/blob/0.13.x/docs/guides/overview.md).

jackvz:
Sure, the webpack script attaches React and Router objects to the window object. You can then use these like this:

```
'use strict';

var Route = Router.Route;

var routes = (
  <Route>
    <Route path='' handler={MyReactComponent}/>
    <Route path='path/:id' handler={MyOtherReactComponent}/>
  </Route>
  );

var RouteHandler = Router.RouteHandler;

Router.run(routes, Router.HashLocation, function(Root) {
  React.render(<Root/>, document.getElementById('content'));
});
```

evanheisler:
Thank you @jackvz!

PR #397, opened 2015-10-02
Binding eslint on javascript files content change

zackexplosion:
When I were reviewing code with eslint, I must using **gulp lint** or **gulp build**  EVERY TIME, if just bindling on the file change event will be better to use.

### DIFF
--- a/app/templates/gulpfile.babel.js
+++ b/app/templates/gulpfile.babel.js
@@ -114,6 +114,7 @@ gulp.task('serve', ['styles', 'fonts'], () => {
   ]).on('change', reload);
 
   gulp.watch('app/styles/**/*.<%= includeSass ? 'scss' : 'css' %>', ['styles']);
+  gulp.watch('app/scripts/**/*.js', ['lint']);
   gulp.watch('app/fonts/**/*', ['fonts']);
   gulp.watch('bower.json', ['wiredep', 'fonts']);
 });


sindresorhus:
That's what the ESLint editor plugins are for.

zackexplosion:
By default, is not binding when you start the local dev server

silvenon:
What @sindresorhus was referring to is e.g. [linter-eslint](https://atom.io/packages/linter-eslint). Plugins for your favorite text editor will work much better than linting on file change, because they lint as you type instead of as you save and you can immediately see the line number, the error etc.

screenshot 2015-10-02 11 18 49

arthurvr:
Honestly I just think we shouldn't do this change to the gulpfile and just recommend people to use those editor plugins.

zackexplosion:
I think eslint plugin is not available on every editor?

silvenon:
Yeah, probably not, but it's still a recommended way to go. You can always add that `gulp.watch` line yourself if your editor doesn't support it. When I'm working in my editor, having to check the terminal for errors every now and then is very inconvenient.

Does your editor have a plugin? Do you have a better solution?

arthurvr:
Just wondering, which editors don't have support for ESLint yet? I guess that's pretty rare or am I wrong?

silvenon:
Closing this because it's a problem editor plugins are much better at solving.

PR #405, opened 2015-10-15
Add note about templating languages

silvenon:
@sindresorhus @arthurvr is this ok? We get a lot of these requests, so I want to make it clear.


sindresorhus:
Text looks ok, but should probably be a `faq.md` document linked to from the readme.

silvenon:
Good idea.

addyosmani:
Can we update this to an FAQ.md?

silvenon:
Done.

PR #413, opened 2015-10-26
Babel support

### LINKED ISSUE #348: Add Babel task/recipe
Hi, would you be open to do an optional babel task (or recipe)? I've got it more or less working but there are some issues with uglify and how to properly fit the useref/serve/dist tasks properly.

neonowy:
Fixes #348.


silvenon:
Awesome! Thanks :+1: 

Like in [generator-webapp](https://github.com/yeoman/generator-webapp), we intended for this feature to be hidden behind a flag because it's so common, i.e. to turn it off you would pass `--no-babel` to `yo gulp-webapp`. Would you it be too much trouble to implement it that way?

Issue #412, opened 2015-10-26
How to debug JavaScript errors?


zarembas:
I think I'm missing something, because I can't really understand how to find errors when the Gulp server crashes. Most of the time, when an error gets thrown, it's in the node.js or react.js file. It's not pointing to the line that actually triggers it. What's the process of debugging when using the generator? Thanks.

silvenon:
Could you be more specific? You have a hard time debugging *both* node and React errors? Do you have an example?

vassyz:
As an example, when running ```gulp build``` I'm getting the following errors:

```
events.js:141
      throw er; // Unhandled 'error' event
      ^
Error
    at new JS_Parse_Error (eval at <anonymous> (/Users/.../client/node_modules/gulp-uglify/node_modules/uglify-js/tools/node.js:22:1), <anonymous>:1508:18)
    at js_error (eval at <anonymous> (/Users/.../client/node_modules/gulp-uglify/node_modules/uglify-js/tools/node.js:22:1), <anonymous>:1516:11)
    at croak (eval at <anonymous> (/Users/.../client/node_modules/gulp-uglify/node_modules/uglify-js/tools/node.js:22:1), <anonymous>:2008:9)
    at token_error (eval at <anonymous> (/Users/.../client/node_modules/gulp-uglify/node_modules/uglify-js/tools/node.js:22:1), <anonymous>:2016:9)
    at unexpected (eval at <anonymous> (/Users/.../client/node_modules/gulp-uglify/node_modules/uglify-js/tools/node.js:22:1), <anonymous>:2022:9)
    at as_property_name (eval at <anonymous> (/Users/.../client/node_modules/gulp-uglify/node_modules/uglify-js/tools/node.js:22:1), <anonymous>:2610:13)
    at eval (eval at <anonymous> (/Users/.../client/node_modules/gulp-uglify/node_modules/uglify-js/tools/node.js:22:1), <anonymous>:2564:24)
    at eval (eval at <anonymous> (/Users/.../client/node_modules/gulp-uglify/node_modules/uglify-js/tools/node.js:22:1), <anonymous>:2055:24)
    at expr_atom (eval at <anonymous> (/Users/.../client/node_modules/gulp-uglify/node_modules/uglify-js/tools/node.js:22:1), <anonymous>:2515:35)
[...]

PR #417, opened 2015-11-12
Make it easier to debug errors in .jsx files

zarembas:
It was very hard for me to understand what the error was from the current output (related to [#412](https://github.com/yeoman/generator-gulp-webapp/issues/412)).
Using console.log() will tell you what the error is, the file that triggers it and the exact line number. The gulp server will also not crash every time it outputs an error. Make sure you keep an eye on the console.

### DIFF
--- a/docs/recipes/react.md
+++ b/docs/recipes/react.md
@@ -6,10 +6,10 @@ This recipe gets you set up with React, including precompilation of JSX into Jav
 
 ### 1. Add dependencies
 
-Install [gulp-react](https://github.com/sindresorhus/gulp-react), a gulp plugin for transforming JSX templates into vanilla JavaScript:
+Install [gulp-babel](https://github.com/babel/gulp-babel), and the required presets for transforming JSX templates into vanilla JavaScript:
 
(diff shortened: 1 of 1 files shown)


silvenon:
Thanks :smiley: Though I think it might be better to replace gulp-react with gulp-babel, see the [deprecation notice](https://github.com/sindresorhus/gulp-react#deprecated-in-favor-of-gulp-babel-read-more). Perhaps gulp-babel has better error handling.

vassyz:
I thought about that as well, but I've never used gulp-babel. This was a quick fix for my current project.

I also added sourcemaps in my second commit.

silvenon:
Should be the same thing, only better. But the error handling should be checked.

addyosmani:
:+1: for gulp-babel

vassyz:
I replaced gulp-react with gulp-babel, and it's working fine. I removed the error checking, as it seems to be built-in into babel. :smirk:

silvenon:
Awesome! Thanks :+1:

Issue #124, opened 2014-05-11
Clean task doesn't run before serve/watch


callumlocke:
The `clean` task is only set as a dep of `default`. I expected it to run before serve/watch also (probably as a dep of `connect`).

Is there any reason for this? If not I'll PR

w0rd-driven:
I may have jumped the gun for creating a pull request for an issue over a year old now with no mention here first. I've had a `clean` dependency using the same del package become a red-headed step-child for a long time before I realized updating the package to 2.x (and syntax) fixed it. Fortunately, it looks to be the same here.

PR #434, opened 2015-12-29
Using del > v2.0.0 corrects #124

w0rd-driven:
The gulp tasks have changed quite a bit since #124 was first introduced but that fortunately made correcting this much easier.

I first verified the problem still persists by adding the `clean` dependency to `serve`. It would fail as expected as either the fonts or styles directories wouldn't be empty during async execution. Once I changed to the promise syntax, `clean` would wait for completion before `fonts` and `styles` began.

I tried to recreate as much of the work done in the issue as possible by also creating a fresh project and testing the changes. I believe both the default `gulp` and `gulp serve` work as expected.

### DIFF
--- a/app/templates/gulpfile.babel.js
+++ b/app/templates/gulpfile.babel.js
@@ -8,7 +8,7 @@ import {stream as wiredep} from 'wiredep';
 const $ = gulpLoadPlugins();
 const reload = browserSync.reload;
 
-gulp.task('styles', () => {<% if (includeSass) { %>
+gulp.task('styles', ['clean'], () => {<% if (includeSass) { %>
   return gulp.src('app/styles/*.scss')
     .pipe($.plumber())
     .pipe($.sourcemaps.init())
@@ -72,7 +72,7 @@ gulp.task('images', () => {
     .pipe(gulp.dest('dist/images'));
 });
 
-gulp.task('fonts', () => {
+gulp.task('fonts', ['clean'], () => {
   return gulp.src(require('main-bower-files')('**/*.{eot,svg,ttf,woff,woff2}', function (err) {})
     .concat('app/fonts/**/*'))
     .pipe(gulp.dest('.tmp/fonts'))
@@ -92,7 +92,7 @@ gulp.task('clean', () => {
   return del.bind(null, ['.tmp', 'dist']);
 });
 
-gulp.task('serve', ['clean', 'styles', 'fonts'], () => {
+gulp.task('serve', ['styles', 'fonts'], () => {
   browserSync({
     notify: false,
     port: 9000,
@@ -165,6 +165,4 @@ gulp.task('build', ['lint', 'html', 'images', 'fonts', 'extras'], () => {
   return gulp.src('dist/**/*').pipe($.size({title: 'build', gzip: true}));
 });
 
-gulp.task('default', ['clean'], () => {
-  gulp.start('build');
-});
+gulp.task('default', ['build']);
--- a/app/templates/gulpfile.babel.js
+++ b/app/templates/gulpfile.babel.js
@@ -88,9 +88,11 @@ gulp.task('extras', () => {
   }).pipe(gulp.dest('dist'));
 });
 
-gulp.task('clean', del.bind(null, ['.tmp', 'dist']));
+gulp.task('clean', () => {
+  return del.bind(null, ['.tmp', 'dist']);
+});
 
-gulp.task('serve', ['styles', 'fonts'], () => {
+gulp.task('serve', ['clean', 'styles', 'fonts'], () => {
   browserSync({
     notify: false,
     port: 9000,
--- a/app/templates/_package.json
+++ b/app/templates/_package.json
@@ -6,7 +6,7 @@
   "devDependencies": {
     "babel-core": "^5.5.6",
     "browser-sync": "^2.2.1",
-    "del": "^1.1.1",
+    "del": "^2.2.0",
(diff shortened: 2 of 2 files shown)


silvenon:
I believe other tasks are run in parallel, just that `clean` task happens to always finish first. Does it work fine even if you don't `return`?

w0rd-driven:
Upgrading del alone doesn't seem to change anything. It still tries to run `clean`, `styles`, and `fonts` in parallel. Using `return` shows the start and finish for `clean` before it moves on.

Before:
```
⋊> /p/p/y/test gulp serve                                               20:44:39
[10:09:16] Requiring external module babel-core/register
[10:09:16] Using gulpfile /projects/personal/yo/test/gulpfile.babel.js
[10:09:16] Starting 'clean'...
[10:09:16] Starting 'styles'...
[10:09:17] Starting 'fonts'...
[10:09:17] 'clean' errored after 1.2 s
[10:09:17] Error: ENOTEMPTY: directory not empty, rmdir '/projects/personal/yo/test/.tmp/fonts'
    at Error (native)
[10:09:17] Finished 'styles' after 1.2 s
[10:09:17] Finished 'fonts' after 854 ms
```

After:
```
⋊> /p/p/y/test gulp serve                                               10:10:10
[10:10:49] Requiring external module babel-core/register
[10:10:49] Using gulpfile /projects/personal/yo/test/gulpfile.babel.js
[10:10:49] Starting 'clean'...
[10:10:49] Finished 'clean' after 204 μs
[10:10:49] Starting 'styles'...
[10:10:49] Starting 'fonts'...
[10:10:50] Finished 'styles' after 932 ms
[10:10:50] Finished 'fonts' after 730 ms
[10:10:50] Starting 'serve'...
[10:10:50] Finished 'serve' after 61 ms
[BS] Access URLs:
 -------------------------------------
       Local: http://localhost:9000
    External: http://192.168.0.12:9000
 -------------------------------------
          UI: http://localhost:3001
 UI External: http://192.168.0.12:3001
 -------------------------------------
[BS] Serving files from: .tmp
[BS] Serving files from: app
```

app/templates/gulpfile.babel.js
```diff
@@ -88,9 +88,11 @@ gulp.task('extras', () => {
   }).pipe(gulp.dest('dist'));
 });
 
-gulp.task('clean', del.bind(null, ['.tmp', 'dist']));
+gulp.task('clean', () => {
+  return del.bind(null, ['.tmp', 'dist']);
```
sindresorhus:
```js
return del(['.tmp', 'dist']);
```

w0rd-driven:
Thanks for that catch. I mistakenly thought it may have been something babel required. I'll amend the PR

w0rd-driven:
... or not. If I take the `.bind()` out, `clean`, `styles`, and `fonts` starts in tandem like the before snippet. `bind()` was added in 90841bb6aa6d3512a7fb293b1727e264efa463e5 to address roughly the same problem. For giggles, I downgraded del back to `^1.1.1` (1.2.1 is installed) and this change still works. It seems that just `() => { return del.bind() }` is the necessary piece that was missing.

sindresorhus:
@w0rd-driven That's not the same situation. With that commit it was incorrect to return as `del` has a callback interface. del@2 has a promise interface and returning is the correct way to do it.

The correct fix here is to make `clean` a dependency on the individual tasks. Tasks are run concurrently, not in series.

silvenon:
Thansk @w0rd-driven for your work on this. Because adding `clean` as a dependency to 4 tasks is still a workaround, I wanted to wait for gulp v4 so we can solve this the right way.

w0rd-driven:
The latest change has `clean` as a dep on `styles` and `fonts`. I personally don't mind that but if those tasks are ran individually, `clean` will be a little too eager.

It seems like through either commit, `bind()` is really the essential piece preventing the tasks from stepping on each other. In the first commit, `clean` placed in any part of the array appears to block and finish before the others execute. By placing it in the front, it would at least produce the behavior I was expecting: `clean` starts, finishes, then the other tasks execute concurrently. In the second commit `bind()` is also necessary so that the `html` and `fonts` deps don't step on each other. `bind()` feels a lot like slight of hand in either scenario and you can make this "work", but they both have a smell to them.


I agree about Gulp4 after feeling like I've exhausted the easier paths. My hope is these commits and my attempts to explain them would help anyone if they chose to implement this downstream.

eddiemonge:
Until Gulp4 is done (or we switch this to it) this isn't the best solution. Its adding a dependency to tasks that don't really need it. Its much easier to just tell people to do `gulp clean && gulp serve`.

However, del could still be upgraded if this PR was just for that.

w0rd-driven:
I thought del was upgraded separately but looking over the commit logs, I guess it wasn't. I could easily do a PR for that though.

For my own edification, what exactly in Gulp4 would fix this in a more elegant way? I guess it's clearly defining the `parallel` vs `series` task dependencies? It's likely that `clean` almost exclusively requires it to be ran in series vs the default parallel in gulp3.

eddiemonge:
right, clean should be ran in a series with all the other tasks ran parallel after it

silvenon:
I have decided to go with run-sequence to fix this problem. It's fixed in 91fe09b78fcb312fe128cc90bc4349d436cd4711 and published in [v2.2.0](https://github.com/yeoman/generator-webapp/releases/tag/v2.2.0). While I was doing the change I forgot about this PR, otherwise I would suggest it here. Thanks for your effort! 👍

PR #440, opened 2016-01-14
Add watch:false to gulp nunjucks task (fixes #406)

### LINKED ISSUE #406: Nunjucks gulp task hangs after build
I added the nunjucks recipe and the gulp tasks are all working fine but the `gulp build` task never completes - it just hangs after finishing everything. 

The site still successfully builds but it seems like the nunjucks views task is causing it not to finish. 

I am using the tasks as written in the recipe.

joehand:


silvenon:
Thanks :+1: I didn't see this PR so I made the change already, but I wanted to merge this to see what will happen. Apparently there are two commits now with the same change? :smile:

PR #456, opened 2016-02-16
Update nunjucks recipe with .njk

joelbschwartz:
The file extension might be a matter of personal preference, but I think using `.nunjucks` brings this closer in line with the Jade recipe, and clarifies that the files are not actually of type `.html`.

`gulp-nunjucks-render` version 2.0.0 now requires the path to templates be configured with a `path` option instead of `nunjucks.configure1` (https://github.com/carlosl/gulp-nunjucks-render#example)

This diverges from the Jade recipe, but naming partials with a leading underscore and putting them in the layouts folder makes inclusion a bit simpler and perhaps keeps the intention of the file clear.

### DIFF
--- a/docs/recipes/nunjucks.md
+++ b/docs/recipes/nunjucks.md
@@ -7,14 +7,14 @@ We assume your directory structure will look something like this:
 ```
 webapp
 └── app
-    ├── about.html
-    ├── contact.html
-    ├── index.html
+    ├── about.njk
+    ├── contact.njk
+    ├── index.njk
     ├── includes
-    │   ├── footer.html
-    │   └── header.html
+    │   ├── footer.njk
+    │   └── header.njk
     └── layouts
-        └── default.html
+        └── default.njk
 ```
 
 If you had something different in mind, modify paths accordingly.
@@ -29,7 +29,7 @@ Install [gulp-nunjucks-render](https://github.com/carlosl/gulp-nunjucks-render) 
 $ npm install --save-dev gulp-nunjucks-render
 ```
 
-### 2. Modify `app/index.html` to create as `app/layouts/default.html` layouts template
+### 2. Modify `app/index.html` to create as `app/layouts/default.njk` layouts template
 
 Modify `app/index.html`:
 
@@ -49,15 +49,15 @@ Modify `app/index.html`:
 Make it the default layout template:
 
 ```
-$ mv app/index.html app/layouts/default.html
+$ mv app/index.html app/layouts/default.njk
 ```
 
-### 3. Create new Nunjucks `app/index.html` page to extend from `app/layouts/default.html`
+### 3. Create new Nunjucks `app/index.njk` page to extend from `app/layouts/default.njk`
 
-Create `app/index.html`:
+Create `app/index.njk`:
 
 ```diff
-+{% extends "layouts/default.html" %}
++{% extends "layouts/default.njk" %}
 +
 +{% block content %}
 +  <div class="hero-unit">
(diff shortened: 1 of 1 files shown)


oleyb:
I'm in full support of ditching the use of the `.html` file extension in favor of `.nunjucks` for nunjucks templates. If we have the ability to accurately convey the content of a file by use of the file extension, why would we not embrace that?

If we like the use of `.html` why not just use `.txt` for all text-based files? That'd make things easier, right? ;)

silvenon:
Although `.nunjucks` is a long extension, I like this idea. I'll wait a bit to see what the others say.

joelbschwartz:
Thanks, @silvenon. I'll look forward to the discussion.

Nunjuck's own documentation uses `.html` in the examples, but they don't comment on it too much. Maybe this is inspired by Jinja, which documents "A Jinja template doesn’t need to have a specific extension".

FWIW from the community perspective: syntax highlighting plugins for  [Atom](https://github.com/alohaas/language-nunjucks/blob/master/grammars/html%20%28nunjucks%20templates%29.cson), [Brackets](https://github.com/axelboc/nunjucks-brackets/), [Sublime](https://github.com/mogga/sublime-nunjucks/blob/master/Nunjucks.tmLanguage), and [Vim](https://github.com/Glench/Vim-Jinja2-Syntax/blob/master/ftdetect/jinja.vim) all enable syntax highlighting for `.nunjucks` (and `.html`), and to a lesser extent `.nunjs`, with single examples for `.nunj` and `.njk`.

I overlooked the comment from @CesarDenis in the Issues that he was going to work on this with #459.

oleyb:
>  syntax highlighting plugins for Atom, Brackets, Sublime, and Vim all enable syntax highlighting for `.nunjucks`

This for me is the kicker; it impacts the workflow of developers. It seems crazy to complicate stuff for devs by using an ambiguous file extension.

joelbschwartz:
@silvenon FYI, the Google Group for Nunjucks seemed like a better forum for this question of file extensions, so I raised the concern there: https://groups.google.com/forum/#!topic/nunjucks/mJbJ96OPUGQ

Perhaps basing this PR on that outcome makes more sense? And using a stacked extension for the template files such as `.html.nj` might be more ideal than `.nj`.

silvenon:
@joelbschwartz I would like to merge this PR with the extension `.njk` (sounds like a better abbreviation). Sorry for the small merge conflict caused by 880d6fd8995b2027d30aab55c4deef01c6bf0042 :sweat_smile:

joelbschwartz:
@silvenon Ok! How do you feel about the stacked extension, `.html.njk`, to identify what the templating language is being used on? This is based on some recent discussion from the Nunjucks Google Group. I am testing now to see if a formula with a stacked extension adversely affects anything.

silvenon:
@joelbschwartz stacked extensions seem mostly like a Rails convention. We should then also use `.css.scss` and `.js.babel`, but I'm not a fan. I would prefer just `.njk` if that's ok.

joelbschwartz:
@silvenon I think it's slightly different for a generic templating engine like Nunjucks because it can be used with _anything_. Whereas `scss` is strictly a `css` extension. The stacked extension clarifies what is being templated. That said, I am fine with using only `.njk`. 😃Updated and ready to go! :+1:

silvenon:
Yeah, let's do it like that for now. One question: I've just seen this change: `path: 'app/layouts'`, doesn't this only allow layouts? What about partials? Maybe it's better to have it set to `app`?

joelbschwartz:
Clearly I've put too much of my own opinion in this PR. 😛My personal preference is to put the partial files directly within the `layouts` folder and identify them with a leading underscore.

```
     └── layouts
         └── _footer.njk
         └── _header.njk
         └── default.njk
```

Or, to add a `_partials` or `partials` subfolder to layouts and put them in there (then, perhaps no underscore on the individual files).

```
     └── layouts
         └── _partials
             └── footer.njk
             └── header.njk
         └── default.njk
```

Happy to change it back to revert to the original `includes` folder and setting path to `app` if that's your preference and/or if there is motivation to make this recipe consistent with the Jade recipe. Let me know, and I'll update the PR accordingly.

silvenon:
I'd prefer to have `layouts` and `includes` side by side, if possible. I feel like that is a convention among static site generators. `_header.njk` is not a layout, it's a partial, so it's confusing to me as well :confused: 

Thanks so much for your help :+1:

joelbschwartz:
K. I just double-checked the diff and I think that's good to go now.

silvenon:
Could you [squash your commits](https://ariejan.net/2011/07/05/git-squash-your-latests-commits-into-one/)? I promise that's the last thing! :wink:

joelbschwartz:
Oops! Squashed. 😀

nukkamoto:
The syntax highlighting for sublime sucks, I can't tell where the template starts and where the actual HTML is.

Syntax highlight is letting me down on this beautiful templating language.

silvenon:
@valtido not sure which plugin you are using, but you could submit a PR to add the `njk` extension. E.g. in sublink-nunjucks that could be [this part](https://github.com/mogga/sublime-nunjucks/blob/master/Nunjucks.tmLanguage#L9-L11) (I don't use Sublime, so I'm not sure).

PR #461, opened 2016-02-17
Update recipe for installing less

eugenious:
I had some difficulty following this recipe because of (recent) changes in the gulpfile, so I've updated it so others hopefully will have an easier time with it.

The diff looks like a dog's breakfast, because some of the +/- are in the diff, and others are actually in the file itself. But I really didn't do that much. Basically what I did was:
1. update the context lines of the changes because they were out of date
2. swapped the order of changes #2 and #3 since they were swapped in the gulpfile
3. some other small changes such as from `main.less` to `*.less`

I've tested this out locally and it seems to work okay, but it might be good to double check it.

Also, and unrelated to this pull request, I believe the section on source maps is no longer needed, because source maps is included by default (at least, it is for me).

Thanks for an awesome generator, I like using it!

### DIFF
--- a/docs/recipes/less.md
+++ b/docs/recipes/less.md
@@ -23,43 +23,42 @@ $ npm uninstall --save-dev gulp-sass && npm install --save-dev gulp-less
 
 ```diff
  gulp.task('styles', () => {
--  return gulp.src('app/styles/main.scss')
--    .pipe($.sass({
--      outputStyle: 'nested',
+-  return gulp.src('app/styles/*.scss')
++  return gulp.src('app/styles/*.less')
+     .pipe($.plumber())
+     .pipe($.sourcemaps.init())
+-    .pipe($.sass.sync({
+-      outputStyle: 'expanded',
 -      precision: 10,
--      includePaths: ['.'],
--      onError: console.error.bind(console, 'Sass error:')
-+  return gulp.src('app/styles/main.less')
+-      includePaths: ['.']
+-    }).on('error', $.sass.logError))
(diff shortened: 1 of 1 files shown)


silvenon:
Thanks for your help :+1:

The `styles` task is still not quite right, though. The diff should look like this:

``` diff
 gulp.task('styles', () => {
   return gulp.src('app/styles/*.scss')
     .pipe($.plumber())
     .pipe($.sourcemaps.init())
-    .pipe($.sass.sync({
-      outputStyle: 'expanded',
-      precision: 10,
-      includePaths: ['.']
-    }).on('error', $.sass.logError))
+    .pipe($.less({
+      paths: ['.']
+    }))
     .pipe($.autoprefixer({browsers: ['> 1%', 'last 2 versions', 'Firefox ESR']}))
     .pipe($.sourcemaps.write())
     .pipe(gulp.dest('.tmp/styles'))
     .pipe(reload({stream: true}));
 });
```

Also, yes, feel free to dump the Source Maps section :+1: We already include them here by not deleting the `$.sourcemaps` part.

When you're done, please [squash your commits](https://ariejan.net/2011/07/05/git-squash-your-latests-commits-into-one/) so you end up with a single one.

eugenious:
Okay, thanks for the feedback.

I pushed an update:
1. fixed the context lines
2. got rid of source maps section
3. squashed the commits, and added a better commit message

Let me know if you need anything else

silvenon:
Thanks! Looks great now :+1:

PR #464, opened 2016-02-19
Add Gulp file include

morrislaptop:
Hey guys,

I thought this would be a useful addition to the generator. It includes the gulp-file-include plugin, which allows people to _optionally_ use code like:

``` html
  @@include('./partials/view.html')
  @@include('./var.html', {
    "name": "haoxin",
    "age": 12345,
    "socials": {
      "fb": "facebook.com/include",
      "tw": "twitter.com/include"
    }
  })
```

This is always useful with small frontend projects where the developer is endlessly copying and pasting snippets of HTML for repeating elements like comments, blog articles, generated content etc... 

If they don't use these snippets, then everything works as it always did :) 

I did rename the html gulp task to useref, and used the html task for gulp file include as I felt it was more relevant, however happy to have these renamed.

### DIFF
--- a/app/templates/_package.json
+++ b/app/templates/_package.json
@@ -14,6 +14,7 @@
     "gulp-cache": "^0.4.2",
     "gulp-cssnano": "^2.0.0",
     "gulp-eslint": "^0.13.2",
+    "gulp-file-include": "^0.13.7",
     "gulp-htmlmin": "^1.3.0",
     "gulp-if": "^1.2.5",
     "gulp-imagemin": "^2.2.1",
--- a/app/templates/gulpfile.babel.js
+++ b/app/templates/gulpfile.babel.js
@@ -37,6 +37,14 @@ gulp.task('scripts', () => {
 });
 <% } -%>
 
+gulp.task('html', function() {
+  return gulp.src('app/*.html')
+    .pipe($.plumber())
+    .pipe($.fileInclude())
+    .pipe(gulp.dest('.tmp/'))
+    .pipe(reload({stream: true}));
+});
+
 function lint(files, options) {
   return () => {
     return gulp.src(files)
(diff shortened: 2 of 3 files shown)


silvenon:
Interesting minimal approach. My three main concerns:
- no syntax highlighting, I assume
- including same things in every page is still repetitive, I prefer having a single layout file which "wraps around" the pages
- as soon as users want something more advanced, they need to switch to something else

I would rather if people went for the [Nunjucks recipe](https://github.com/yeoman/generator-webapp/blob/880d6fd8995b2027d30aab55c4deef01c6bf0042/docs/recipes/nunjucks.md), it still seems fairly simple.

morrislaptop:
Hey,
- No syntax highlighting
- Layout files - I guess this is a matter of preference however I would say that a layout file requires a lot of extra planning and forethought - whereas I'm using this generator to hit the ground running
- Something more advanced - sure, this can always happen and in this instance they could follow the Nunjucks recipe

Cheers,
Craig

silvenon:
I'd definitely like to have some built-in templating solution, because it has been requested so many times. I'll take some time to make the decision.

morrislaptop:
Great! Let me know if I can do anything else :) 

Looping in @coderhaoxin (author of gulp file include) if he has any suggestions on the implementation.

morrislaptop:
- bump!

musdy:
+1

silvenon:
We decided [not](https://github.com/yeoman/generator-webapp/blob/master/docs/faq.md#add-templating-engine-x) to have a built-in templating solution, they are too arbitrary without a clear winner. 😕 Would you like to add gulp-file-include as a recipe?

morrislaptop:
Cool, I gave nunjucks a go and it was definitely a better workflow then gulp file include

PR #470, opened 2016-02-22
Delete images from dist folder as they are deleted from the app folder.

NigelOToole:
Added clean-images task which watches for image deletions in the app/images folder and deletes corresponding image in the dist/images folder images.


silvenon:
Thanks for your contribution :+1: However I don't see how this is necessary, the `dist` directory is cleaned up before each build and it doesn't serve a purpose in development.

Could you elaborate?

NigelOToole:
My mistake, I was using `gulp build` rather than just `gulp` to build my project. Thanks for taking a look.

Issue #467, opened 2016-02-20
gulp command fails. 


joepegler:
Getting this for a fresh installation on Ubuntu 15.10. 

`/home/xx/node_modules/gulp-util/lib/PluginError.js:73
  if (!this.message) throw new Error('Missing error message');
                           ^
Error: Missing error message
`

[http://i.imgur.com/s7jsm1k.png]

joepegler:
'gulp styles' is the culprit...

cvasquez:
Having the same issue here. Would appreciate any help.

joepegler:
I had to roll back the previous version that uses grunt.

silvenon:
Which Node version are you running? Try using v4 or later.

If that doesn't work, this is a stretch, but try using [gulp-postcss](https://github.com/postcss/gulp-postcss) + [autoprefixer](https://github.com/postcss/autoprefixer) instead of gulp-autoprefixer. So instead of this:

``` js
// in the styles task
.pipe($.autoprefixer({browsers: ['> 1%', 'last 2 versions', 'Firefox ESR']}))
```

write this:

``` js
// along with other imports
import autoprefixer from 'autoprefixer';
```

``` js
// in the styles task
.pipe($.postcss([
  autoprefixer({browsers: ['> 1%', 'last 2 versions', 'Firefox ESR']})
]))
```

sassquad:
I tried the above also, but this fails. Using node v0.10.42 and npm v3.7.3 on Mac OS X El Capitan:

```
TypeError: Object #<Object> has no method 'postcss' at 
Gulp._gulp2.default.task._gulp2.default.src.pipe.pipe.pipe.pipe.pipe.pipe.reload.stream 
(gulpfile.babel.js:22:13)
```

Having used nvm to install node v4.3.1 and switch accordingly, the build now proceeds without incident.

cvasquez:
Thanks so much @sassquad - node 4.3.1 worked for me!

silvenon:
@sassquad before trying the 2nd version you need to install `gulp-postcss` and `autoprefixer`. But I'm glad you got it working with a newer version of Node, that was the problem then :+1: 

I'll reopen if that doesn't fix it for @joepegler.

sassquad:
@cvasquez - Glad to be of help!

@silvenon - I'll look into your suggestion when I get near my computer later.

sassquad:
[...]

PR #471, opened 2016-02-22
Update to yo-gen 0.22 and update tests

eddiemonge:
- Removes entries from gitignore for old style tests
- Update to use yeoman-test
- Update tests to run in os tmp dir instead of test/temp

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,1 @@
 node_modules
-test/gulp-file
-test/sass
-test/slug
-temp/
--- a/package.json
+++ b/package.json
@@ -29,15 +29,16 @@
   "dependencies": {
     "chalk": "^1.0.0",
     "generator-jasmine": "^0.2.1",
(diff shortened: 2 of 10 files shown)


silvenon:
I'd like to dump Node 0.x versions, our generator doesn't seem to work on those anyway. #467 

All aboard? @sindresorhus @SBoudrias @addyosmani

sindresorhus:
:+1:

SBoudrias:
:+1: - good update btw, I love all that red!

silvenon:
I was hoping pushing to master will apply to Travis tests for this PR, apparently I need GitHub lessons. Could you cherry pick 3e4526f548f7195658dc771e3e41e94f84458419 or something?

eddiemonge:
changed the one es2015 feature I was using so it should pass in older tests and won't be considered a breaking change

PR #473, opened 2016-02-26
Disable cssnano autoprefixer

chitchu:
Cssnano autoprefixes by default. Which can override your gulp-autoprefixer options.


silvenon:
Oh my. Thank you :+1:

PR #485, opened 2016-03-17
Update screenshot

fgilio:
Update screenshot with new 'generator-webapp' and 'webapp' strings in title bar, also update to OS X El Capitan look.


sindresorhus:
Thanks :)

PR #486, opened 2016-03-20
Update jade docs follow version 2.0.0

nguyenmanh1507:
Update content in gulpfile.babel.js follow ver 2.0.0 and add gulp-plumber to prevent views task break.


silvenon:
Thanks!

PR #487, opened 2016-03-23
Update links in jade recipe because bootstrap-sass-official is bootst…

iamgalba:
Hey guys, path to bootstrap-sass in jade recipe needs to be updated in step 5


silvenon:
Thanks!

PR #496, opened 2016-04-05
Update gh-pages.md

BronzePointKrel:
Recipe was written with gulp-clean which is not in package.json by default.

### DIFF
--- a/docs/recipes/gh-pages.md
+++ b/docs/recipes/gh-pages.md
@@ -12,10 +12,10 @@ Deploying your app using `git subtree` allows the use of [GitHub Pages](https://
 
 > You can check your remotes with `$ git remote -v`. See [Adding a remote](https://help.github.com/articles/adding-a-remote) for more info.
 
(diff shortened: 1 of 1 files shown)


silvenon:
This line will error out then, no?

``` js
.pipe($.clean());
```

I'm not sure why it's cleaning after deployment anyway. But it should clean up before, which is not happening currently. Seems to me something like this is needed:

``` js
gulp.task('upload', ['build'], () => {
  // gh pages stuff
});

gulp.task('deploy', ['clean'], () => {
  gulp.start('upload');
});
```

Which gets around the issue where Gulp cannot specify the order in which tasks will run.

BronzePointKrel:
Yes, without the change it produces an error.

The clean is needed because this recipe requires you to remove 'dist' from .gitignore 

The last step removes dist so that it isn't listed as an untracked folder and the entire build folder is not committed.

silvenon:
Oh, my bad. I thought you were removing `gulp-clean` from the recipe, not adding it. Nevermind :grin:

PR #500, opened 2016-04-09
Add build dependency before serving built app

ghost:
Create a fresh, distributable copy of the app before serving

### DIFF
--- a/app/templates/gulpfile.js
+++ b/app/templates/gulpfile.js
@@ -135,7 +135,7 @@ gulp.task('serve', ['styles', 'fonts'], () => {
   gulp.watch('bower.json', ['wiredep', 'fonts']);
 });
 
-gulp.task('serve:dist', () => {
+gulp.task('serve:dist', ['build'], () => {
   browserSync({
     notify: false,
     port: 9000,


silvenon:
Thanks for your contribution 👍 I was thinking about this as well, but imagine how frustrating it would be having that task rebuild the app every time. Maybe you want to shut down the server for some reason, then later check again, now you have to rebuild again. If you have a good reason for this that I overlooked, I'm open to convincing.

kyleloopback:
For me personally, I served a dist copy, shut down the server, made some changes, and ran the command again, only to find my changes weren't there. Seems odd to have to explicitly build your app before serving the built copy. I found _that_ frustrating. No worries if you don't add it in to the gen though, easy add on my side.

ghost:
For me personally, I served a dist copy, shut down the server, made some changes, and ran the command again, only to find my changes weren't there. Seems odd to have to explicitly build your app before serving the built copy. I found _that_ frustrating. No worries if you don't add it in to the gen though, easy add on my side.

silvenon:
Hm, you're right. The main reason why you'd want to restart `serve:dist` is probably to check new changes. Could you rebase and resolve conflicts? Then I'll merge it.

sindresorhus:
ping @alwynsays

ghost:
Sorry about the major lag! Rebased and ready to go.

silvenon:
Thanks! I changed it to run `default` instead of `build` to ensure a clean build.

PR #515, opened 2016-05-09
remove unnecessary reload in nunjucks recipe

oddui:
this also matches current `gulpfile.babel.js`

### DIFF
--- a/docs/recipes/nunjucks.md
+++ b/docs/recipes/nunjucks.md
@@ -177,10 +177,10 @@ Edit your `serve` task to watch HTML files in `.tmp`, and so that (a) editing an
       'app/images/**/*'
     ]).on('change', reload);
 
-+   gulp.watch('app/**/*.html', ['views', reload]);
-+   gulp.watch('app/**/*.njk', ['views', reload]);
-    gulp.watch('app/styles/**/*.scss', ['styles', reload]);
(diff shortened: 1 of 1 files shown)


silvenon:
Thanks!

Trost123:
Right now if you have more than one .html generated, you get multiple calls to browsersync (one for each generated html).
So if my website consists of 5 pages, every time I edit one, browsersync will refresh 5 times for each html generated.

What fixed it for me:
1)

```
gulp.watch([
  .tmp/*.html //Remove this
]).on('change', reload);
```

2)
`gulp.watch('app/**/*.njk', ['nunjucks', reload]); //Add the reload here`

Also: 
Why are we watching:
`gulp.watch('app/**/*.html', ['views']);`
Makes no sense to me.

silvenon:
@Trost123 good point. I agree with the 1st step, but the 2nd step may be problematic because `nunjucks` and `reload` are executing at the same time, it's just a coincidence that `nunjucks` finishes before browser reload. We should add it at the bottom of the `nunjucks` task, like we do in the [Pug recipe](https://github.com/yeoman/generator-webapp/blob/master/docs/recipes/pug.md#2-create-a-views-task).

Would you like to submit a PR?

Trost123:
Sure.
PR: #583

PR #526, opened 2016-05-23
Bump `gulp-imagemin`

kevva:
If we want to keep the options we can do:

``` js
$.imagemin([
    $.imagemin.gifsicle({interlaced: true}),
    $.imagemin.jpegtran({progressive: true}),
    $.imagemin.optipng(),
    $.imagemin.svgo({plugins: [{cleanupIDs: false}]})
]);
```


silvenon:
Thanks Kevin!

PR #540, opened 2016-06-22
update scaffolding command

bhuber2010:


bhuber2010:
I guess my install was holding onto the old name

PR #562, opened 2016-08-24
Add wiredep to build's dependencies

jechol:
Currently, `bower.json` is only applied to html when watching.
Better to apply it during `build` also.


silvenon:
Thanks for your PR 👍 The problem is that `wiredep` might not be run first, all these tasks are run in parallel, which is a limitation of Gulp v3. We should think of a solution which ensures that `wiredep` will be run first, maybe as a dependency of the `default` task?

sindresorhus:
Closing for lack of feedback from OP.

PR #565, opened 2016-08-29
Recipe Handlebars: Approach to compile into a single file, using partials, minifiy and export to dist

eBaeza:
This is a new approach where the templates are compiled into a single file, also i added a task to handle partials and finally i added the right way to append the compiled files in the `index.html` and that these are exported to `dist`directory

### DIFF
--- a/docs/recipes/handlebars.md
+++ b/docs/recipes/handlebars.md
@@ -10,7 +10,7 @@ This recipe shows how to set up Handlebars to precompile your templates, includi
 Install some gulp plugins:
 
 ```
-$ npm install --save-dev gulp-handlebars gulp-define-module gulp-declare
+$ npm install --save-dev path gulp-concat gulp-handlebars gulp-define-module gulp-declare
 ```
 
 * [gulp-handlebars](https://github.com/lazd/gulp-handlebars) precompiles raw `.hbs` templates into JavaScript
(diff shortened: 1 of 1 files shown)


docs/recipes/handlebars.md
```diff
@@ -23,33 +23,72 @@ $ bower install --save handlebars
...
+    "devDependencies": {
+      ...
+    },
+    "overrides": {
+      "handlebars": {
+        "main": ["handlebars.runtime.js"]
+      }
+    }
+  }
+  ```
```
silvenon, line 42: These code block is not rendering correctly, could you fix that?

docs/recipes/handlebars.md
```diff
@@ -23,33 +23,72 @@ $ bower install --save handlebars
...
+        }
+      }
+    }))
+    .pipe($.concat('partials.js'))
+    .pipe(gulp.dest('.tmp/templates'));
+});
+```
+This compiles the `.hbs` files inside the `templates/partials` directory into `partials.js` file in the `.tmp/templates` directory.
 
 ### 3. Add `templates` as a dependency of both `html` and `serve`
```
silvenon, line 83: This title needs to include the `partials` task (so "Add `templates` and `partials`…").

docs/recipes/handlebars.md
```diff
@@ -58,32 +97,39 @@ gulp.task('serve', ['styles', 'templates', 'fonts'], () => {
 Edit your `serve` task so that (a) editing an `.hbs` file triggers the `templates` task, and (b) the browser is reloaded whenever a `.js` file is generated in `.tmp/templates`:
 
 ```diff
- gulp.task('serve', ['styles', 'templates', 'fonts'], () => {
+gulp.task('serve', ['styles', 'templates', 'partials', 'scripts', 'fonts'], () => {
```
silvenon, line 100: The original `gulpfile.js` has changed (that is my fault for not merging this earlier, sorry 😓). Could you please update this task to reflect the changes? So instead of adding `partials` and `scripts` to the task itself, add it to the [line below](https://github.com/yeoman/generator-webapp/blob/master/app/templates/gulpfile.js#L97).

docs/recipes/handlebars.md
```diff
@@ -58,32 +97,39 @@ gulp.task('serve', ['styles', 'templates', 'fonts'], () => {
...
-<script src="bower_components/handlebars/handlebars.runtime.js"></script>
-<script src="templates/foo.js"></script>
-<script src="scripts/main.js"></script>
+  <!-- build:js scripts/vendor.js -->
+  <!-- bower:js -->
+  ...
+  <!-- endbower -->
+  <!-- endbuild -->
+
+  <!-- build:js templates/index.js -->
```
silvenon, line 129: Maybe I'm missing something here, but why open another comment block, why not just merge templates and partials with the `scripts/main.js` block next to it?

docs/recipes/handlebars.md
```diff
@@ -10,7 +10,7 @@ This recipe shows how to set up Handlebars to precompile your templates, includi
 Install some gulp plugins:
 
 ```
-$ npm install --save-dev gulp-handlebars gulp-define-module gulp-declare
+$ npm install --save-dev path gulp-concat gulp-handlebars gulp-define-module gulp-declare
```
silvenon, line 13: I believe gulp-wrap is missing here? Also, `path` is a built-in module, so I don't think it should be there.

silvenon (changes requested):
Awesome PR! I left some comments, after resolving those I'll merge. 👍

silvenon:
@eBaeza I'm sorry that I'm so late with the review, please let me know if there's anything else I can help with.

PR #566, opened 2016-08-30
Copy fonts to .tmp/ or dist/ depending on task.

Trost123:
Right now changing fonts while running serve or build task will copy fonts to both "dist/fonts" and ".tmp/fonts", which doesn't seem logical to me. Development tasks shouldn't modify the dist, and vice-versa.

Used gulp-if to choose either ".tmp\fonts" (for serve task) or "dist\fonts" (for build task) as a destination for fonts.
I've added a variable "dev= false" which is set to false (via 'set-prod-mode' task) to true on "build".
Not sure if it's a good practice, can someone confirm?

I'm also considering another change to 'fonts' task, using [gulp-newer](https://www.npmjs.com/package/gulp-newer), to skip re-writing unmodified fonts, saving the HDD from useless work.
This can be also done for images.
Should I do it?

### DIFF
--- a/app/templates/gulpfile.js
+++ b/app/templates/gulpfile.js
@@ -8,6 +8,11 @@ const wiredep = require('wiredep').stream;
 const $ = gulpLoadPlugins();
 const reload = browserSync.reload;
 
+var dev = true;
+gulp.task('set-prod-mode', function () {
+  dev = false;
+});
+
 gulp.task('styles', () => {<% if (includeSass) { %>
   return gulp.src('app/styles/*.scss')
     .pipe($.plumber())
@@ -87,8 +92,7 @@ gulp.task('images', () => {
 gulp.task('fonts', () => {
   return gulp.src(require('main-bower-files')('**/*.{eot,svg,ttf,woff,woff2}', function (err) {})
     .concat('app/fonts/**/*'))
-    .pipe(gulp.dest('.tmp/fonts'))
-    .pipe(gulp.dest('dist/fonts'));
+    .pipe($.if(dev, gulp.dest('.tmp/fonts'), gulp.dest('dist/fonts')));
 });
 
 gulp.task('extras', () => {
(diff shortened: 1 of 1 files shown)


silvenon:
Hey 😃 The problem is that there's no guarantee that `set-prod-mode` will be executed before the `fonts` task, because these tasks are executed in parallel. Maybe add `set-prod-mode` as the dependency of the `default` task?

Trost123:
Tested it in my projects - works as expected.
But I'll push the dependency fix, just to make sure it's 100% fail-safe.

P.s. I can also contribute implementations of: 
1. Gulp-critical (for inlining above-the fold css into html).
2. Gulp-if (to skip copying of fonts to .tmp, if they already exist there).
3. Gulp.spritesmith (sprite generation).
4. Copying of bower image assets (that are listed in .bower main) to dist/images.
4. Run-sequence (to do 'wiredep' before 'default').

Should I add some of them to this (or seperate) PR, maybe add some of them to recipes?
(Or maybe these are too situational, and you don't need them at all?)

silvenon:
Maybe there's a way to this without creating the `set-prod-mode` task just to set a variable. Have you tried setting the variable in the `default` task instead?

``` js
gulp.task('default', ['clean'], () => {
  dev = false;
  gulp.start('build');
});
```

Thanks for you contribution ideas! 👍 I'd be interested in running `wiredep` before `serve` and `default` in a separate PR. I think for `default` there's no need for run-sequence, just add it as a dependency next to `clean`, but `serve` is a bit tricky. You can open a PR with the suggested solution and we can discuss from there. 😉

Generating image sprites and extracting critical CSS are great ideas for recipes. 👍

Trost123:
I'm not sure is it notifies you about commits, so I'm bumping this.

silvenon:
@Trost123 would you like to send another PR for this if you still want this change? I'm so sorry for the delay.

Trost123:
I based this PR on my master branch  :-1: , but then I needed to push into master, so I closed this PR.
Another PR: #585

PR #576, opened 2016-09-27
Added Docker deployment support

sbehrends:
Added Dockerfile for deployment.
Thought is not necessary for development, since you just used yeoman.

### DIFF
--- a/app/templates/.dockerignore
+++ b/app/templates/.dockerignore
@@ -1,0 +1,2 @@
+node_modules
+bower_components
--- a/app/templates/Dockerfile
+++ b/app/templates/Dockerfile
@@ -1,0 +1,24 @@
+FROM sbehrends/docker-webapp
+MAINTAINER Sergio Behrends
+
+# Create app directory
+RUN mkdir -p /usr/src/app
+WORKDIR /usr/src/app
+
+# Install node dependencies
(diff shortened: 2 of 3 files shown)


app/templates/.dockerignore
```diff
@@ -0,0 +1,2 @@
+node_modules
+bower_components
```
neilhem, line 2: Should be an empty newline at the end of file

app/templates/Dockerfile
```diff
@@ -0,0 +1,24 @@
...
+
+# Bundle app source
+COPY . /usr/src/app
+
+# Build app
+RUN gulp build
+
+# Serve app
+EXPOSE 9000
+CMD [ "gulp", "serve:dist" ]
```
neilhem, line 24: Should be an empty newline at the end of file

neilhem:
Hi, thank you for your request. Would you mind to add commit with fixes, so we will be able to merge your PR?

sindresorhus:
I think this makes more sense as a recipe than an option. We're not opinionated on how you distribute the built app.

silvenon:
I agree with @sindresorhus, it would be great as a recipe!

sbehrends:
Thanks for the feedback. I will create a recipe then and submit a new PR.

PR #583, opened 2016-10-30
Fixed multi-reload in Nunjucks recipe.

Trost123:
Multiple BrowserSync reload calls would happen, if you have more than one page generated by Nunjucks.

### DIFF
--- a/docs/recipes/nunjucks.md
+++ b/docs/recipes/nunjucks.md
@@ -82,6 +82,7 @@ gulp.task('views', () => {
       path: 'app'
     }))
     .pipe(gulp.dest('.tmp'))
+    .pipe(reload({stream: true}));
 });
 ```
 
@@ -164,19 +165,11 @@ Wiredep does not support `.njk`, so also add in the file type definition.
 
 ### 9. Edit your `serve` task
 
-Edit your `serve` task to watch HTML files in `.tmp`, and so that (a) editing an `app/**/*.html` or `app/**/*.njk` file triggers the `views` task, and (b) reloads the browser:
+Edit your `serve` task so that editing an `app/**/*.html` or `app/**/*.njk` file triggers the `views` task:
 
 ```diff
   gulp.task('serve', ['views', 'styles', 'fonts'], () => {
     ...
-    gulp.watch([
-      'app/*.html',
-+     '.tmp/*.html',
-      '.tmp/styles/**/*.css',
-      'app/scripts/**/*.js',
-      'app/images/**/*'
-    ]).on('change', reload);
-
 +   gulp.watch('app/**/*.html', ['views']);
 +   gulp.watch('app/**/*.njk', ['views']);
     gulp.watch('app/styles/**/*.scss', ['styles']);


silvenon:
Excellent, thanks so much! There's just two things:
1. Why does changing an HTML file trigger `views`?
2. This way changing an HTML file in `app` no longer triggers a page refresh, no?

Trost123:
You can include HTML files in .njk files:
https://mozilla.github.io/nunjucks/templating.html#include
So if you change an HTML, that is included, Nunjucks needs to compile again.

Not sure if that's a common practice though. It was in the recipe when I found this repo, but I never used this.
I actually removed this watcher in my fork:
`gulp.watch('app/**/*.html', ['views']);`
I just keep all partials as .njk and only refresh browser if HTML is changed.

So here we have 2 options:
1)  Only refresh on HTML change, compile .njk on .njk change:

```
gulp.watch([
      'app/*.html', //watch for html's that don't use nunjucks, reload
      'app/images/**/*',
      '.tmp/fonts/**/*'
    ]).on('change', reload);

  gulp.watch('app/**/*.njk', ['views']); //watch for nunjucks
```

2) For those who include html in .njk:

```
gulp.watch([
    //removed original html watcher
      'app/images/**/*',
      '.tmp/fonts/**/*'
    ]).on('change', reload);

  gulp.watch('app/**/*.html', ['views']); //compile njk on html change, reload is included in views task
  gulp.watch('app/**/*.njk', ['views']);
```

Trost123:
Also, right now if you change any .njk, all html files get rewritten, no matter what you change.
I've fixed it in my fork:

```
gulp.task('nunjucks', () => {
  return gulp.src('app/*.njk')
    .pipe($.plumber()) //Added plumber
    .pipe($.nunjucksRender({
      path: 'app'
    }))
    .pipe($.cached('njkCache')) //Cache rendered html, drop unchanged HTML out of the pipe
    .pipe($.print(function(filepath) {
      return "njk compiled: " + filepath; //Logs into console, which files changed
    }))
    .pipe(gulp.dest('.tmp')) //write changed files
    .pipe($.htmlhint()) // (optional) htmlhint changed files
    .pipe($.htmlhint.reporter())
    .pipe(reload({stream: true})); //reload browser
});
```

This way if you change index.njk - only index.html is updated, if you change header.njk - all pages that include/import header.njk will get updated.
Not sure if it's a good thing, or an over-optimizing.

silvenon:
Got it, I didn't know you can add HTML files to Nunjucks. Thanks so much for your contribution. 👍

Trost123:
I tried using this again and found out it still doesn't work properly.
.pipe(reload({stream: true})); inside "views" task gets triggered too early for some unknown reason, so the user has to manually refresh the browser 2nd time too see the latest changes.
Example: original_file.html>change1>save file>change2>save file, my browser will show the following: original_file>autoreload>original_file>autoreload>change1.

Ironically, the approach which you criticized for being unreliable in (#515) works like a charm for me though:
gulp.watch('app/**/*{html,njk}', ['views', reload]);
I have multiple njk files depending on each other built, and the refresh is never triggered too early.
It's weird. This approach is wrong theoretically (because it's synchronous), but it works.
I'm not sure what to do (PR-wise).

silvenon:
@Trost123 sure! Would you like to "correct" it by submitting a PR?

PR #585, opened 2016-11-01
Choose fonts destination with gulp-if.

Trost123:
We only copy to '.tmp/fonts' or 'dist/fonts', depending on task run.


silvenon:
Thanks a bunch! 👍

PR #596, opened 2016-11-08
Removed `options` param from lint function

radum:
After this code [commit](https://github.com/yeoman/generator-webapp/commit/796efb90a442f390fa689a776c5f7bab527af537) we no longer need the `options` param when calling the `lint` function.


silvenon:
That is correct, thanks! 👍

Issue #603, opened 2016-11-17
Ubuntu 14 / 16 "unmet peer dependancy yo@>=1.7.1


750430:
Hi all.

I'm using Ubuntu 14, but have experienced the same issues in v16 too.

If I do:
`jonathan@ubuntu14:~$ npm ls -g yo
/home/jonathan/.npm-global/lib
└── yo@1.8.5 
`
It's evident that Yeoman is installed.

If I run:
`jonathan@ubuntu14:~$` npm install --global generator-webapp` I see:

`npm WARN deprecated minimatch@2.0.10: Please update to minimatch 3.0.2 or higher to avoid a RegExp DoS issue
/home/jonathan/.npm-global/lib
├── generator-webapp@2.3.2 
└── UNMET PEER DEPENDENCY yo@>=1.7.1`

And querying minimatch:
`jonathan@ubuntu14:~$ npm ls -g minimatch
/home/jonathan/.npm-global/lib
├─┬ generator-webapp@2.3.2
│ ├─┬ generator-jasmine@0.2.2
│ │ └─┬ yeoman-generator@0.20.3
│ │   └─┬ findup-sync@0.2.1
│ │     └─┬ glob@4.3.5
│ │       └── minimatch@2.0.10 
│ └─┬ wiredep@4.0.0
│   └─┬ glob@7.1.1
│     └── minimatch@3.0.3 
├─┬ gulp-cli@1.2.2
│ └─┬ matchdep@1.0.1
│   └─┬ findup-sync@0.3.0
│     └─┬ glob@5.0.15
│       └── minimatch@3.0.3 
├─┬ jshint@2.9.4
│ └── minimatch@3.0.3 
├── minimatch@3.0.3 
└─┬ yo@1.8.5
  └─┬ yeoman-doctor@2.1.0
    └─┬ twig@0.8.9
      └── minimatch@3.0.3 
`

I guess there's something going on with paths.

This is a freshly installed Ubuntu VM and I'm trying to follow the guide here [http://yeoman.io/learning/index.html](url)

I had to fix issues with permissions, so followed the guide here: [https://docs.npmjs.com/getting-started/fixing-npm-permissions](url) and used option 2 as the path was simply set to /usr

Any help on getting this resolved would be greatly appreciated.
Regards.

neilhem:
Hi, so you are not able to run `yo webapp`?

mischah:
Hej @750430,

[...]

PR #606, opened 2016-11-29
Remove yo as peer dependency

mischah:
Reason: avoid confusion.

Beside #603 there are questions from time to time in the Yeoman Gitter channel.

### DIFF
--- a/package.json
+++ b/package.json
@@ -26,9 +26,6 @@
     "jquery",
     "gulp"
   ],
-  "peerDependencies": {
-    "yo": ">=1.7.1"
-  },
   "dependencies": {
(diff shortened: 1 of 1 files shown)


silvenon:
But users should have yo installed, no? The fact that npm is sometimes not registering yo to be installed is another problem…

mischah:
> But users should have yo installed, no? 

Yeah. They have to. That’s the reason why npm3 warns about unmet peer dependencies. 
But people doesn’t seem to understand that kinda cryptic warning.

silvenon:
Yep, you're right. `peerDependencies` make more sense for non-global dependencies.

PR #608, opened 2016-12-09
Update README

CyberPunkCodes:
Adds "yo" to install command to fix "UNMET PEER DEPENDENCY" errors.

### DIFF
--- a/README.md
+++ b/README.md
@@ -31,7 +31,7 @@ If your favorite feature is missing and you really need Ruby Sass, you can alway
 ## Getting Started
 
 - Install dependencies: `npm install --global yo gulp-cli bower`
-- Install the generator: `npm install --global generator-webapp`
+- Install the generator: `npm install --global yo generator-webapp`
 - Run `yo webapp` to scaffold your webapp
 - Run `gulp serve` to preview and watch for changes
 - Run `bower install --save <package>` to install frontend dependencies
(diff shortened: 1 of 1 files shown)


silvenon:
@WadeShuler thanks for the contribution! 👍 Isn't the line **right above** saying to install `yo` globally?

> - Install dependencies: `npm install --global yo gulp-cli bower`

silvenon:
Oh, I just saw your explanation in another issue. Instead of double-installing it like this, I think we should tell users to ignore the warning. What do you think?

CyberPunkCodes:
No, after review, it should be how I have illustrated in the updated pull request.

If you look at other generators, they list the dependencies all together in one combined command. Example: [https://github.com/yeoman/generator-angular](https://github.com/yeoman/generator-angular)

This seems to be an issue on OSX (which is a large percent of the community). While it seems redundant, this is how they do it. I believe the pulled packages are cached, so repeating say `gulp cli` or `yo` in the install command, doesn't add too much extra overhead.

No, you should not tell people to ignore it, especially when you can prevent it by an appropriate install command in your README's install instructions. Telling people to just ignore it, will likely lead to more issues being opened here about the same issue. People just don't read :)

eddiemonge:
What referenced issue are you talking about?

silvenon (approved):
I like your update much better. Now there's no confusing duplication and it works around the bug. Thanks!

PR #616, opened 2016-12-27
Nunjucks recipe, browsersync call fix. (again)

Trost123:
Moved the refresh call to "serve" task from "views".
As discussed in #583


silvenon:
Thanks so much! 👍

Trost123:
Oops! I think this PR was not a good solution.
Now I hit this issue:
https://github.com/BrowserSync/browser-sync/issues/717
Basically, the guys who developed BS don't recommend using
`gulp.watch(jsSrc, ['js', browserSync.reload]); // no longer recommended`
as it causes some issues. (you can read the details there in their issue)

Maybe we should revert this one. Sorry about the PR spam.

silvenon:
Done. Thanks! 👍

PR #622, opened 2017-01-23
s3-deploy recipe back to working order

jkatsnelson:
The recipe was using 'build' instead of 'default' which resulted in no fonts being uploaded.

Also, the gulp plugin now uses 'params' instead of 'bucket' directly.

### DIFF
--- a/docs/recipes/aws-s3-deployment.md
+++ b/docs/recipes/aws-s3-deployment.md
@@ -18,12 +18,12 @@ $ npm install --save-dev gulp-awspublish
 Add this task to your `gulpfile.js`. It will run `build` task before deploying:
 
 ```js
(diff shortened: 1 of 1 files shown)


silvenon:
Thanks!

jkatsnelson:
I left out the fact that I removed access key and secret from this recipe, because I didn't want it to seem like that was a good idea.

silvenon:
Maybe we could advise using a gitignored `.env` file + [dotenv](https://github.com/motdotla/dotenv).

jkatsnelson:
If you download the aws-cli, and add your credentials to that, you don't even need .env + dotenv -- it'll just work seamlessly :)

PR #623, opened 2017-02-02
fix RegEx issue within html task

jeffreysbrother:
The generator in its current state is not successfully uglifying JS or minifying CSS/HTML after running `gulp` and then `gulp serve:dist`. I believe this is because the Regular Expressions in the Gulp html task are mistakenly included as strings. 

Additionally, I see no use for the "$" character ... or the "b" character in the cssnano task. Perhaps there is a use for these, but the solution I've provided seems to work.

**Edit**: I intended to also exclude the "$" in this commit, but did not. Sorry about that!


silvenon:
Thanks!

silvenon:
I will release a version soon.

PR #624, opened 2017-02-04
Missing space in index.html

stophecom:


stophecom:
This is huge!

silvenon:
Thanks! Can the space be before `class` (inside the `if` statement), rather than after `h1`? That way we won't have an extra space if `legacyBootstrap` is `true`.

neilhem (commented):
Please add space inside if statement

stophecom:
@neilhem makes sense

PR #629, opened 2017-02-13
Added Liquify recipe

imalitavakoli:
Liquify is a plugin which rendres Liquid template language inside of our
webapp project.


silvenon:
Wow, thanks! 👍

PR #631, opened 2017-02-15
Nunjucks + Browsersync recipe fix v3

Trost123:
Force gulp to wait for .njk files to compile before triggering Browsersync reload.
According to this guide: https://browsersync.io/docs/gulp#gulp-reload

P.S. Took me a lot of tries to deal with this one, while the solution was in Browsersync docs all along 😃

### DIFF
--- a/docs/recipes/nunjucks.md
+++ b/docs/recipes/nunjucks.md
@@ -72,10 +72,10 @@ gulp.task('views', () => {
 
 This compiles `app/*.njk` files into static `.html` files in the `.tmp` directory.
 
-### 5. Create a 'views_reload' task
+### 5. Create a 'views:reload' task
 
 ```js
-gulp.task('views_reload', ['views'], () => {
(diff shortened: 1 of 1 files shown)


docs/recipes/nunjucks.md
```diff
@@ -72,7 +72,17 @@ gulp.task('views', () => {
 
 This compiles `app/*.njk` files into static `.html` files in the `.tmp` directory.
 
-### 5. Add `views` as a dependency of both `html` and `serve`
+### 5. Create a 'views_reload' task
```
silvenon: I would prefer that we use `:` for separating task words instead of `_`. Could you change this (and the following occurrences) to `views:reload`?

silvenon (changes requested):
Thanks a lot for continuously refining this recipe!

silvenon (approved):

PR #675, opened 2017-06-19
Added webpack support

tsndr:


silvenon:
@tsndr congratulations on your first PR! Could you give more context for this PR? What problem does it solve?

tsndr:
There ist no problem to solve, this PR only adds webpack support, which enables you to include javascript files using `require("script.js")`

silvenon:
@tsndr how does that play with the rest of the build system? For example, why would we still use useref for JS if we can include all JS files with `require`?

tsndr:
It was just an idea and in my projects it works fine, so i thought, that it would be a nice feature for the webapp package

silvenon:
Sorry, I didn't mean to discourage you. I really love webpack and I would like to integrate it with generator-webapp, I'm just not sure to what extent this works and what is recommended usage. Would we instruct users to only use this for local files? Also, if we first apply Babel **then** webpack, could we get ES6 `import` syntax?

I get confused when Gulp and webpack mix. I usually use one or the other.

Issue #674, opened 2017-06-16
Fonts not moving to dist folder on build


jamilalisgandarov:
None of the fonts are moving from app/fonts to dist/fonts properly. Not even fonts from bower-components. Here is the task from gulpfile.js. I have not changed anything that webapp has included.
`gulp.task('fonts', () => {
  return gulp.src(require('main-bower-files')('**/*.{eot,svg,ttf,woff,woff2}', function (err) {})
    .concat('app/fonts/**/*'))
    .pipe($.if(dev, gulp.dest('.tmp/fonts'), gulp.dest('dist/fonts')));
});`

akshayvinchurkar1:
your gulp task is fine. any errors in the build process?

jamilalisgandarov:
Nope, no errors.

IniterWorker:
Create a new task with unset boolean **dev mode**

```javascript
gulp.task('dev-false', () => {
  dev = false;
});
```

Add a new task in the order of the build task

```javascript
gulp.task('build', ['dev-false', 'lint', 'html', 'images', 'fonts', 'extras'], () => {
  return gulp.src('dist/**/*').pipe($.size({title: 'build', gzip: true}));
 });
```

CesarDenis:
In my case are the fonts from the bower-components. I'm using [shim repository for Font Awesome](https://github.com/components/font-awesome) not moving to dist or .tmp folder.

IniterWorker:
I read an comment in this case. Because, Awesome and Bootstrap don't give correctly the font directory path in the configuration of the main-bower-files. @CesarDenis 

I try with bootstrap-sass and awesome, no problem.

```json
"dependencies": {
    "bootstrap-sass": "~3.3.5",
    "font-awesome": "fontawesome#^4.7.0",
  }
```

CesarDenis:
@IniterWorker I tried and had the same problem.
```
"dependencies": {
    "bootstrap": "~4.0.0-alpha.6",
    "font-awesome": "fontawesome#^4.7.0"
}
```
- node: `v6.11.0`
- bower: `v1.8.0`
- gulp: `v3.9.1`

IniterWorker:
@CesarDenis. My mistake.

The issue of copying is really solved. But my exemple use `bootstrap-sass`. The `fontawesome` tricked me with CDN in my app. She don't work. I'll try to explain why.

The main-bower-file return
```json
[...]

PR #681, opened 2017-07-11
#674 newTask change boolean dev

IniterWorker:
There is no means of changing the boolean dev mode.
I make a new task to handle this. #674

### DIFF
--- a/app/templates/gulpfile.js
+++ b/app/templates/gulpfile.js
@@ -11,6 +11,10 @@ const reload = browserSync.reload;
 
 let dev = true;
 
+gulp.task('dev-false', () => {
+  dev = false;
+});
+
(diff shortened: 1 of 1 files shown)


silvenon (commented):
It's being set to default in the `default` task (which is at the bottom). Why do you need to set it in `build` as well?

IniterWorker:
@silvenon because I'm a bad boy and i don't see the default task with dev = false. I see build. I execute the build task. This practice may be familiar for me ("build task").

silvenon:
The documentation says to run `gulp`, not `gulp build`. `gulp` also cleans and runs wiredep, so that's the task you should use.

Thanks for being so kind to contribute, let me know if everything is clear and if I should reopen this.

PR #687, opened 2017-08-13
Updated to bootstrap~4.0.0-beta

MarkyMarkNo1:
... which requires popper.js: https://github.com/FezVrasta/popper.js#installation 

Unfortunately it seems impossible to override bootstraps' dependencies in bower.json 😣

### DIFF
--- a/app/index.js
+++ b/app/index.js
@@ -170,7 +170,8 @@ module.exports = class extends Generator {
 
       // Bootstrap 4
       bowerJson.dependencies = {
-        'bootstrap': '~4.0.0-alpha.6'
+        'bootstrap': '~4.0.0-beta',
+        'popper.js-unpkg': 'https://unpkg.com/popper.js'
       };
 
       // Bootstrap 3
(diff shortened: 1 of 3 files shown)


silvenon:
Thanks @MarkyMarkNo1! What are you trying to override? Does the [`overrides` field](https://roots.io/using-bootstrap-with-bower-how-to-override-bower-packages/) not work?

MarkyMarkNo1:
My pleasure, @silvenon ;)

Bootstrap comes with a popper.js dependency. 

Since a popper.js dist is not available via Bower - see the link I gave you above or take a look inside what Bower downloads - we need the unpkg.com/popper.js workaround.

Hence we should just override Bootstraps' bower `dependencies` in our bower.json - but that doesn't seem to work. We can override `main`, but not `dependencies`.

We'd need something like 

```
overrides: {
  "bootstrap": {
     "dependencies": {
         "jquery": ">=1.9.1",
         "popper.js": "https://unpkg.com/popper.js"
      }
  }     
}

```

silvenon:
Ok, to me this is fine because Bower is deprecated anyway, we can't expect a nice solution because of its limitations.

partyparrotgreg:
Hey guys, I think it's time to update Bootstrap to 4.1.0. Switching version bower.json cause some issues and it doesn't work.

PR #706, opened 2018-01-02
temporary solution for the fonts task

smakosh:
the fonts don't get copied after building but this temporary solution works fine!

### DIFF
--- a/app/templates/gulpfile.js
+++ b/app/templates/gulpfile.js
@@ -89,10 +89,12 @@ gulp.task('images', () => {
     .pipe(gulp.dest('dist/images'));
 });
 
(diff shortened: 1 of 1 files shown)


Stormix:
+1

UlisesGascon:
I was trying to fix the conflict and maybe we have solved this bug already. Can you recheck it @smakosh ?
https://github.com/yeoman/generator-webapp/blob/master/app/templates/gulpfile.js#L127

silvenon:
After the gulp v4 update this no longer seems to be the issue. Thanks @smakosh for your effort to fix this! ❤️

PR #730, opened 2018-10-22
Add ability to configure server port

### LINKED ISSUE #729: Can we use process.env.PORT, too?
I use environments like c9.io and most of the time `9000` is not the right port for me....

My idea is to add support for `process.env.PORT`  ;-)

I will submit a PR asap as version `3.0.2`.

UlisesGascon:
Closed #729

### DIFF
--- a/app/templates/gulpfile.js
+++ b/app/templates/gulpfile.js
@@ -11,7 +11,9 @@ const cssnano = require('cssnano');
 const $ = gulpLoadPlugins();
 const reload = browserSync.reload;
 
+
 let dev = true;
+const port = process.env.PORT || 9000;
 
 gulp.task('styles', () => {<% if (includeSass) { %>
   return gulp.src('app/styles/*.scss')
--- a/app/templates/gulpfile.js
+++ b/app/templates/gulpfile.js
@@ -7,7 +7,7 @@ const wiredep = require('wiredep').stream;
 const runSequence = require('run-sequence');
 const autoprefixer = require('autoprefixer');
 const cssnano = require('cssnano');
-const argv = require('yargs').argv;
+const { argv } = require('yargs');
 
 const $ = gulpLoadPlugins();
 const reload = browserSync.reload;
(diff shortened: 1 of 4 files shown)


silvenon (commented):
Yeah, we should definitely have this configureable. I was thinking that we do this using parameters instead of environment variables.

```
gulp serve --port 8000
```

You can use yargs for this, see [this Stack Overflow answer](https://stackoverflow.com/questions/28538918/pass-parameter-to-gulp-task). With environment variables we would have to use cross-env, so it's an extra dependency either way.

What do you think?

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "generator-webapp",
-  "version": "3.0.1",
+  "version": "3.0.2",
```
silvenon: Don't bump the version, I will do it, instead just let me know that you would like to see this feature published ASAP. 😉

silvenon (changes requested):

package.json
UlisesGascon: Oh! Great! I will update it now ^^

UlisesGascon:
Hi @silvenon! I added yargs to the project.
- `gulp serve`, `gulp serve:test` and `gulp serve:dist` now accept additional param `--port=xxxx`., by default we use 9000.
- I added some examples to the `Readme.md` too...

app/templates/gulpfile.js
```diff
@@ -7,11 +7,13 @@ const wiredep = require('wiredep').stream;
 const runSequence = require('run-sequence');
 const autoprefixer = require('autoprefixer');
 const cssnano = require('cssnano');
+const argv = require('yargs').argv;
```
silvenon: So sorry to nitpick, but you can destructure this instead:
```suggestion
const { argv } = require('yargs');
```

silvenon (approved):
Looks great, thanks for updating the docs too! After updating what I left in the comments, I'll merge.

app/templates/gulpfile.js
UlisesGascon: Right! Let me change that 👍

UlisesGascon:
Awesome! First PR almost done 👍

Issue #736, opened 2018-11-06
Modernizr build limitations


UlisesGascon:
I am working on the migration from Bower #700 and I am using npm to install modernizr. This will require an additional compilation, based on the [current documentation](https://www.npmjs.com/package/modernizr). 

**We have several options:**
- Not include Modernizr anymore
- Use another source like cdnjs or similar to get the valid build version
- Add an additional step in the generation to perform this build
- Install the dependency and let the final user to perform the build, using [Terminal option](https://www.npmjs.com/package/modernizr#to-generate-everything-in-config-alljson) or [method exposed](https://www.npmjs.com/package/modernizr#building)
- More options... ?

What do you think?

silvenon:
This would probably be the "right" way to do it:

- add the `modernizr` dependency
- add a `modernizr.json` file containing a couple of example options from [config-all.json](https://github.com/Modernizr/Modernizr/blob/master/lib/config-all.json)
- somehow ensure that `gulp serve` and `gulp` trigger Modernizr to build using `modernizr.json` as configuration

Maybe this would be a good time to create two npm scripts, `start` and `build`, which would contain something like:

- `"start": "modernizr -c modernizr.json && gulp serve"`
- `"build": "modernizr -c modernizr.json && gulp"`

I'm just brainstorming here. I think including all of Modernizr would be kinda wasteful, but on the other hand allowing people to fine-tune the build might be a lot of work from our side… What do you think?

UlisesGascon:
yeah! Let me explore this option. I think NPM Scripts is the best solution even for the future in order to keep the same way of working ^^. Thanks for your support!

silvenon:
No problem, thanks for fixing all these problems!

I just found out about [gulp-modernizr](https://www.npmjs.com/package/gulp-modernizr), which seems like the best fit for this project.

silvenon:
Actually, AFAICT, it doesn't detect CSS. I want people to be able to use selectors like `.no-touchevents .app {}`, but I don't think that gulp-modernizr is able to discover that and build appropriately.

UlisesGascon:
Yeah! I have some issues with modernizer....
I had the idea to build it from node_modules but there are some problems with the routes, `./node_modules/modernizr/bin/modernizr -c modernizr.json -d app/scripts/modernizr.js` 

```
[...]

Issue #700, opened 2017-11-27
New Update without Bower?


dertuerke:
Hi,

Is a new update planned without Bower?

neilhem:
Hi, we are going to release a release without Bower, but there are some blocking issues

cesarferreira:
what's wrong with bower, not cool anymore? Moving to `npm` ?

neilhem:
Bower authors announced deprecation of bower and as an alternative using npm/yarn.

san983:
Hi @neilhem is there a thread for this? https://github.com/yeoman/generator-webapp/issues/700#issuecomment-347492542 

I'm willing to help, thanks!

xied75:
Any updates? Is this still in plan?

spetrey:
Is still in the pipeline? I'm assuming there are still some major blocking issues. How can we help?

silvenon:
@smpetrey thanks for offering to help! A PR containing following changes would be immensely appreciated:

## Update all Yeoman dependencies

Those would probably these three:

  - `yeoman-generator`
  - `generator-mocha`
  - `generator-jasmine`

## Replace Bower and wiredep with npm

This means that from now on users will have to link dependencies manually from `node_modules` manually, instead of automatically from `bower_components`.

## Tests

Ensuring that every combination of options returns an expected new output would be great.

UlisesGascon:
Hi @smpetrey & @silvenon! Can I help here?

silvenon:
Sure @UlisesGascon! This didn't really take off, so if you'd like to take on this upgrade, I'd be more than happy!

UlisesGascon:
@silvenon  Sure! I will start soon ;-)

UlisesGascon:
@silvenon do you want to create a dedicated branch, like `legacy-bower`? I think that this change will be a major one xD

silvenon:
I'm not sure. Bower is deprecated so I don't know if keeping a legacy branch is necessary. Are you asking because we might potentially backport some fixes to that branch?

UlisesGascon:
Oh! Well... I saw [legacy-grunt](https://github.com/yeoman/generator-webapp/tree/legacy-grunt), but it is ok for me to keep the same branches as we have today ^^

silvenon:
The difference is that Grunt isn't deprecated.

UlisesGascon:
One more question... Bower will be replaced with NPM, that is an easy change.

[...]

PR #737, opened 2018-11-07
Migration from Bower to NPM

UlisesGascon:
### Current Status
- [x]  Let's ditch support for Bootstrap v3 🔥
- [x] Remove Bower/Wiredep/etc...
    - [x] Migrate Bower dependencies to NPM (BS3, BS4)
    - [x] Rewrite `app/templates/index.html`, etc...
    - [x] Update gulp tasks (remove wiredep/bower related ones and add new ones of needed)
    - [x] Remove unnecessary files
- [x] Remove Bower references from Documentation (`app/USAGE`,  `README.md`, etc...)
- [ ] Remove Bower references from Recipes
- [x] Update Tests

### Know issues (to be fixed in this PR)
- [x] [Bug with Modernizr Build, reported by @JelmarBosscher](https://github.com/yeoman/generator-webapp/issues/736#issuecomment-436875951)

### Open Discussions (please, Join us!)
- [ ] #700 What is the next step for this generator regarding module bundle concerns?

### Important!

**I will review all the comments/suggestions/reviews... As this is a major change, many things can go wrong, so let's try to prevent that together! :-)**

I will let open this PR until I finish all the steps above. Then I will ask for the final approval/review

### Notes

- [x] I am aware that testings are falling.
- [x] I am aware that some combinations can crash the generator.
- [x] I just started to modify the docs first and then I will jump to update the tests.
- [x] I am working on that too.


README.md
```diff
@@ -30,23 +29,22 @@ If your favorite feature is missing and you really need Ruby Sass, you can alway
...
 - Run `yo webapp` to scaffold your webapp
-- Run `gulp serve` to preview and watch for changes
-- Run `gulp serve --port=8080` to preview and watch for changes in port `8080`
-- Run `bower install --save <package>` to install frontend dependencies
-- Run `gulp serve:test` to run the tests in the browser
-- Run `gulp serve:test --port=8085` to run the tests in the browser in port `8085`
-- Run `gulp` to build your webapp for production
-- Run `gulp serve:dist` to preview the production build
-- Run `gulp serve:dist --port=5000` to preview the production build in port `5000`
+- Run `npm run serve` or `npm start` to preview and watch for changes
```
silvenon: I think it's confusing to have two scripts doing the same thing, I suggest having only `npm start` and deleting the `serve` script.

app/templates/modernizr.json
```diff
@@ -0,0 +1,289 @@
...
+    "websockets/binary",
+    "window/framed",
+    "workers/blobworkers",
+    "workers/dataworkers",
+    "workers/sharedworkers",
+    "workers/transferables",
+    "workers/webworkers",
+    "xdomainrequest"
+  ]
+}
```
silvenon, line 27: I suggest not adding all feature detects, instead just a few as an example, because people won't need most of them. I suggest leaving these, they are quite common:

```
css/supports
img/sizes
img/srcset
serviceworker
touchevents
```

I wonder if it's possible to add a comment inside this file linking to the entire list 🤔 

Also, add newline at the end of the file.

JelmarBosscher:
Forked your feature branch and will use it for an experimental project. 
Any issues I'll encounter will be put in a comment here for now.

P.S. Thank you for getting rid of bower.

JelmarBosscher:
(I know this issue is a big work in progress, just going to put stuff here to keep you on track)

After using your 700 feature branch to build a new project with Bootstrap, Sass and Modernizr the `npm start` command wasn't working. More about that here: #736 

After skipping the modernizr command and manually using the gulp task `gulp serve`
I noticed because of bower being gone you decided to put the following in the app/index.html.
```
<link rel="stylesheet" href="/node_modules/bootstrap/dist/css/bootstrap.min.css" type="text/css"/>
``` 
The same goes for 
```
 <script type="text/javascript" src="/node_modules/jquery/dist/jquery.min.js"></script>
 <script type="text/javascript" src="/node_modules/bootstrap/dist/js/bootstrap.min.js"></script> 
```

This of course didn't go well when I used the task `gulp build`, it still had the path to the node_modules.

If there is going to be a direct path to bootstrap in the node_modules for the time being. 
We can easily resolve this by doing the following.

_app/templates/main.scss_
```
<% if (includeBootstrap) { %>
  // @import 'variables'; /* Overwrite the default Bootstrap variables */
  @import './node_modules/bootstrap/scss/bootstrap';
<% } -%>
```
This will both allow the user to change the bootstrap variables before being processed and bootstrap will be included in to the minified main.css. 

For the javascript files there needs to be a better solution.

silvenon:
@JelmarBosscher could you clarify what's not good about the JS solution?

README.md
```diff
@@ -57,7 +55,7 @@ If your favorite feature is missing and you really need Ruby Sass, you can alway
 - `--skip-install-message`
   Skips the the message displayed after scaffolding has finished and before the dependencies are being installed.
 - `--skip-install`
-  Skips the automatic execution of `bower` and `npm` after scaffolding has finished.
+  Skips the automatic execution of `npm` after scaffolding has finished.
```
silvenon: This would make it more clear and tool-independent:

```suggestion
  Doesn't automatically install dependencies after scaffolding has finished.
```

For example, we also support Yarn, which isn't included here, and we shouldn't have to think about that when writing the docs.

app/index.js
```diff
@@ -75,18 +72,18 @@ module.exports = class extends Generator {
         name: 'Bootstrap 4',
         value: false
       }],
-      when: answers => answers.features.indexOf('includeBootstrap') !== -1
+      when: answers => answers.features.includes('includeBootstrap')
```
silvenon: Oooh, nice ✨

app/index.js
```diff
@@ -377,42 +248,10 @@ module.exports = class extends Generator {
...
-    if (this.includeSass) {
-      // wire Bower packages to .scss
-      wiredep({
-        bowerJson: bowerJson,
-        directory: 'bower_components',
-        ignorePath: /^(\.\.\/)+/,
-        src: 'app/styles/*.scss'
-      });
-    }
-  }
```
silvenon, line 417: 🔥🔥🔥

app/templates/_package.json
```diff
@@ -3,6 +3,22 @@
...
+    "modernizr": "^3.6.0",
+    <% } %>
+    <% if(includeJQuery && !includeBootstrap) { %>
+    "jquery": "^3.3.1"
+    <% } %>
+    <% if (includeBootstrap && !legacyBootstrap) { %>
+    "bootstrap": "^4.1.3",
+    "jquery": "^3.3.1"
+    <% } %>
+    <% if(legacyBootstrap) { %>
```
silvenon: Bootstrap v4 is now stable (it was in alpha when we added support from it), let's ditch support for v3. 🔥

app/templates/_package.json
```diff
@@ -39,12 +55,27 @@
...
-    "main-bower-files": "^2.5.0",
     "mocha": "^5.2.0",
     "run-sequence": "^2.2.1",
-    "wiredep": "^4.0.0",
     "yargs": "12.0.2"
-  },
+  }, 
+  "scripts": {
+    <% if (includeModernizr) { %>
+    "build:modernizr": "./node_modules/modernizr/bin/modernizr -c modernizr.json -d app/scripts/modernizr.js",
```
silvenon: I suggest renaming this script to `modernizr` instead because it's not a variation of the `build` script, like `serve:dist` is a variation of `serve`. (I won't include this in the suggestion below because it has to be edited in multiple places.)

Also, within npm scripts you don't have to worry about targeting local executables, that behavior is built-in:

```suggestion
    "build:modernizr": "modernizr -c modernizr.json -d app/scripts/modernizr.js",
```

UlisesGascon:
Thanks again for checking the changes @silvenon & @JelmarBosscher! It is super nice to have this feedback now :-)

Maybe... I got your point @JelmarBosscher with the `node_modules` paths... When we used Bower and Wiredep we managed to create a build tasks that bundle all the dependencies together. `/script/vendors...` and `/script/app.min.js`. Without Bower and Wiredep we need to implement this manually, until we will use a new application module bundler such as Webpack or Parcel. 

So... not for this PR, but soon the idea is to implement one of this bundles in order to concat & minify all the js and css into one file, right @silvenon ? 

We started a discussion about that in #700

app/templates/gulpfile.js
```diff
@@ -179,31 +178,17 @@ gulp.task('serve:test', () => {
...
 });
 
 gulp.task('default', () => {
   return new Promise(resolve => {
     dev = false;
-    runSequence(['clean', 'wiredep'], 'build', resolve);
+    runSequence('clean', 'build', 'build:size', resolve);
   });
-});
+});
```
silvenon, line 192: I suggest configuring your text editor to recognize `.editorconfig` files, that way you'll avoid diffs like this. We require newline at the end of each file.

app/templates/gulpfile.js
```diff
@@ -179,31 +178,17 @@ gulp.task('serve:test', () => {
...
-<% } %>
-  gulp.src('app/*.html')
-    .pipe(wiredep({<% if (includeBootstrap) { if (includeSass) { %>
-      exclude: ['bootstrap<% if (legacyBootstrap) { %>-sass<% } %>'],<% } else { %>
-      exclude: ['bootstrap.js'],<% }} %>
-      ignorePath: /^(\.\.\/)*\.\./
-    }))
-    .pipe(gulp.dest('app'));
-});
+gulp.task("build:size", () => {
```
silvenon: If this task only measures the file size and doesn't do anything to the files, I suggest renaming it to `size` instead.

app/templates/index.html
```diff
@@ -9,18 +9,16 @@
     <link rel="apple-touch-icon" href="apple-touch-icon.png">
     <!-- Place favicon.ico in the root directory -->
 
-    <!-- build:css styles/vendor.css -->
-    <!-- bower:css -->
-    <!-- endbower -->
-    <!-- endbuild -->
```
silvenon, line 15: Don't remove `build` comments, they're from gulp-useref, which concatenates and compresses JS and CSS. Or are you using another approach that I didn't notice?

You are correct to remove `bower` comments because they're from wiredep

app/templates/modernizr.json
```diff
@@ -0,0 +1,289 @@
+{
+  "minify": true,
```
silvenon: Perhaps minifying is not necessary because we'll do that with gulp-useref in the build step. Leaving it unminified may help when debugging.

silvenon (changes requested):

silvenon:
Concerning @JelmarBosscher's comment, the only problem I see is removing `build` comments, which I pointed out in the review.

JelmarBosscher:
It was a bit of a late night comment and you did manage to understand what I meant. :+1:

Your discussion at #700 does show you guys decided to leave the node_modules in the index.html.
Eventually Webpack or Parcel would resolve the things I mentioned.

UlisesGascon:
Thanks for all the feedback and the code review @silvenon! I will update the PR with the last suggestions and then I will focus on minors bugs and testing :-^)

README.md
UlisesGascon: I agree!

app/index.js
UlisesGascon: Thanks!

app/index.js
UlisesGascon: yeah! :metal::metal::metal:

app/templates/modernizr.json
UlisesGascon: 👌

app/templates/index.html
UlisesGascon: My mistake! I wasn't aware of this build way before, I will reinclude this lines and JS related too

app/templates/modernizr.json
UlisesGascon: great! I will update the list asap!

app/templates/_package.json
UlisesGascon: 🤘 yeah!! Let me implement this one in a separate commit, just to keep all more clear in the changelog ^^

app/templates/gulpfile.js
UlisesGascon: Oh! let me see what can I do... right now I am using c9.io... and the support for `.editorconfig` has been [delayed](https://community.c9.io/t/support-editorconfig/5813) for a long time...

app/templates/gulpfile.js
UlisesGascon: I fixed this one too ^^

app/templates/_package.json
UlisesGascon: 🤘Done! https://github.com/yeoman/generator-webapp/pull/737/commits/e912dacef54ea5347766ee02cd832d36e904ca5a

UlisesGascon:
@JelmarBosscher the bug related to modernizr should be fixed now.. please can you confirm?
@silvenon Please.. check my last changes and the refactor/addition of the tests

I think that we are getting closer to finish with the legacy ^^

app/templates/index.html
```diff
@@ -3,24 +3,23 @@
...
-
     <!-- build:css styles/main.css -->
     <link rel="stylesheet" href="styles/main.css">
     <!-- endbuild -->
     <% if (includeModernizr) { %>
-    <!-- build:js scripts/vendor/modernizr.js -->
-    <script src="/bower_components/modernizr/modernizr.js"></script>
-    <!-- endbuild --><% } %>
+    <script src="scripts/modernizr.js"></script>
```
silvenon, line 15: Is this going to be uglified in production? If not, I think we should keep the `build:js` comments.

app/templates/index.html
```diff
@@ -90,17 +89,13 @@ <h1>'Allo, 'Allo!</h1>
...
     </script>
-
-    <!-- build:js scripts/vendor.js -->
-    <!-- bower:js -->
-    <!-- endbower -->
-    <!-- endbuild -->
+    <!-- build:js scripts/vendor.js -->	
+    <!-- endbuild -->	
+    <% if (includeJQuery || includeBootstrap) { %>
+      <script type="text/javascript" src="/node_modules/jquery/dist/jquery.min.js"></script>
```
silvenon, line 92: I think jQuery should be within `build.js` comments. Even though this one is already compressed, useref _concatenates_ all vendor scripts into one `vendor.js` file, resulting in a single HTTP request, which is what we're aiming for.

app/templates/index.html
```diff
@@ -90,17 +89,13 @@ <h1>'Allo, 'Allo!</h1>
...
+    <% } %>
     <% if (includeBootstrap) { %>
-    <!-- build:js scripts/plugins.js -->
-<% bsPlugins.forEach(function (plugin) { -%>
-    <script src="<%= bsPath + plugin %>.js"></script>
-<% }) -%>
-    <!-- endbuild -->
+      <script type="text/javascript" src="/node_modules/bootstrap/dist/js/bootstrap.min.js"></script>
```
silvenon, line 95: I think it's best to place both jQuery and Bootstrap under the `vendor.js` build comment.

silvenon (commented):
Sorry for replying so late. Really amazing work @UlisesGascon!

I left a few more comments regarding production.

app/templates/index.html
UlisesGascon: I agree! :-)

app/templates/index.html
UlisesGascon: Yes! Sounds fine to me

app/templates/index.html
UlisesGascon: God idea! Let me change that now ^^

JelmarBosscher:
> @JelmarBosscher the bug related to modernizr should be fixed now.. please can you confirm?
> @silvenon Please.. check my last changes and the refactor/addition of the tests
> 
> I think that we are getting closer to finish with the legacy ^^

@UlisesGascon 
After running `npm start` modernizr.js is placed within app/scripts and correctly linked to in the index.html.
So **YES** the issue I mentioned earlier is fixed.

**But**

`gulp serve` on the other hand does not create a modernizr.js in the scripts folder.
Obvious error: `Failed to load resource: the server responded with a status of 404 (Not Found)`
The documentation could be changed to have the user start with `npm start` else there will not be a modernizr in the app or dist folder.

**Or** 

my suggestion would be to use [npm postinstall](https://docs.npmjs.com/misc/scripts) and have modernizr.js be placed in app/scripts after npm install has ran.
All is needed is to add `"postinstall": "npm run build:modernizr",`. in here [app/index.js:59](https://github.com/UlisesGascon/generator-webapp/blob/7004e839730a1291c6bf8a62a9b98cfd22f105d9/app/templates/_package.json#L59) 

Like that it does not matter what command the user starts with and modernizr.js will be in place either way.
Even if the user decides to use the the `--skip-install` option and use `npm install` themselves later on.

---
I tried the above suggestion locally and it works out well.
Though I'm not sure why it states `undefined` in the following and what effect is has.
```
> undefined postinstall H:\Personal\web-app
> npm run build:modernizr


> @ build:modernizr H:\Personal\web-app
> modernizr -c modernizr.json -d app/scripts/modernizr.js
```

silvenon:
@JelmarBosscher but the documentation **is** changed to tell users to use `npm start` instead. Is there a place in the docs where there is still `gulp serve`?

app/templates/index.html
```diff
@@ -90,18 +86,17 @@ <h1>'Allo, 'Allo!</h1>
...
       ga('create','UA-XXXXX-X');ga('send','pageview');
     </script>
-
-    <!-- build:js scripts/vendor.js -->
-    <!-- bower:js -->
-    <!-- endbower -->
-    <!-- endbuild -->
+    <!-- build:js scripts/vendor.js -->	
+    <% if (includeModernizr) { %>
+    <script src="scripts/modernizr.js"></script>
```
silvenon: Please keep Modernizr in `<head>`, otherwise there might be a flash of outdated content, because Modernizr can affect CSS too, which means that the page will first load the styles **before** Modernizr was loaded.

You can surround it in build comments with the name `scripts/vendor/modernizr.js` like it was before.

silvenon (commented):
@UlisesGascon thanks! One last thing and that's it!

JelmarBosscher:
@silvenon You are correct, I may still have had the older readme file locally.
But wouldn't it still be better for modernizr to be build after npm install? 
What was the reasoning for it to be build everytime you want to gulp serve?

Edit: I assume so that people can config their own modernizr.json? 
If so my above mentioned suggestion is indeed invalid :pensive:

Could we just implement the postinstall either way for the users that may not use the documented commands?

app/templates/index.html
UlisesGascon: I restored it 👍

app/templates/_package.json
```diff
@@ -3,6 +3,22 @@
   "engines": {
     "node": ">=4"
   },
+  "dependencies": {
+    <% if (includeModernizr) { %>
+    "modernizr": "^3.6.0",
```
JelmarBosscher, line 11: This line will still have a comma `,` when I scaffold without jQuery and bootstrap. 
Will throw `npm ERR! JSON.parse Failed to parse json`.

silvenon:
`postinstall` only builds when you install dependencies, what if you change the `modernizr.json` file? You'll still get the old Modernizr script until you install dependencies again. It doesn't seem like the appropriate event for this.

app/templates/_package.json
silvenon: Oh wow, good catch!

UlisesGascon: Let me fix that! 👍

app/templates/index.html
silvenon: This is what I meant:

```html
<!-- build:js scripts/vendor/modernizr.js -->
<script src="scripts/modernizr.js"></script>
<!-- endbuild -->
```

This way it will be compressed in production.

UlisesGascon: Now should be fine here https://github.com/yeoman/generator-webapp/pull/737/commits/f9ccfa817a58e43497423f16d1acb7eab2e4e013

silvenon (approved):

silvenon:
LGTM! Is this PR ready on your end? Should I merge it? I remove the WIP text from the description.

UlisesGascon:
We can merge it if you want ^^

But I have some tasks pending related to Docs, like `Remove Bower references from Documentation (recipes, app/USAGE, README.md, etc...`

silvenon:
Oh, right! Perhaps it's best to solve them in this PR then. 😄

UlisesGascon:
I updated the main documentation files. I will work now in the recipes, last step  :muscle::muscle:

UlisesGascon:
@silvenon Do you think that we should keep this PR for `/recipes` too? Maybe we can have a better look and discussion bout them. I am thinking that some things are outdated on top of Bower dependencies. 

If you agree.. we can merge this PR and I will open a new issue asap 👍

silvenon:
Yes, let's discuss recipes in a new issue/PR 😃 

Huge thanks for this!

marekdano:
When do you think this change will be released and updated on NPM? I'd love to use this generator without bower.

silvenon:
I'd love to give an ETA, but I can't commit to it because I'm very busy. 😕 GitHub now has a feature to send you a notification when a new version has been released in a repo, you can use it.

alienriquebm:
So, on the next release, how can we add an dependence with npm/yarn and use it in the project? I'm so happy for your hard work!

silvenon:
Well, you'll have to manually add paths to dependencies in `node_modules`. Wiredep used to automatically inject `<link>`, `<script>` etc., but it only works for Bower, so it might actually feel like a bit of a downgrade until we implement a bundler like Parcel.

The long term upside is that we will no longer be using a deprecated package manager and module bundling might really improve developer experience.

Issue #740, opened 2018-12-01
We need some refactor


UlisesGascon:
I was using Code Climate and Codacy in order to understand better our source code. I will perform fast refactor in order to reduce the complexity and bad things, manly in `app/index.js`.

[This is my starting point](https://codeclimate.com/github/UlisesGascon/generator-webapp/issues?category=complexity&engine_name%5B%5D=structure&engine_name%5B%5D=duplication). I will send a PR asap 👍 

Maybe @silvenon... can grant access to this tool so we can add this checks in the PRs using Travis if you like this idea ^^

silvenon:
Never used Code Climate before, I guess we could add it, I haven't seen it around much lately, though. I was thinking of introducing ESLint + Prettier or something.

UlisesGascon:
Yeah! I can work on that too... 👍

silvenon:
Fixed with #741.

PR #739, opened 2018-11-30
Google Analytics is now Optional

UlisesGascon:
This was discussed in #735 

### Features
- [x] Code updated
- [x] New tests added for this feature

PR #741, opened 2018-12-01
Refactor in order to improve maintainability

UlisesGascon:
### Related to #740 

📝  **[Main Report](https://codeclimate.com/github/UlisesGascon/generator-webapp/issues?severity%5B%5D=major&severity%5B%5D=minor&category%5B%5D=complexity&status%5B%5D=&status%5B%5D=open&status%5B%5D=confirmed&status%5B%5D=invalid&status%5B%5D=wontfix&engine_name%5B%5D=structure&engine_name%5B%5D=duplication&language_name%5B%5D=JavaScript)**

**Critical points covered in this refactor**
- [x] File `index.js` has 365 lines of code (exceeds 250 allowed)
- [x] Function `_writingBower` has 83 lines of code (exceeds 25 allowed). 
- [x] Function `_writingHtml` has 52 lines of code (exceeds 25 allowed). 
- [x] Function `prompting` has 48 lines of code (exceeds 25 allowed). 
- [x] Function _writingHtml has a Cognitive Complexity of 7 (exceeds 5 allowed).
- [x] Most of the features can be setup now in `./lib/config.js`
- [x] New abstraction in order to generate and copy files.


app/index.js
```diff
@@ -84,146 +42,39 @@ module.exports = class extends Generator {
...
-    this.fs.copy(
-      this.templatePath('favicon.ico'),
-      this.destinationPath('app/favicon.ico')
-    );
-
-    this.fs.copy(
-      this.templatePath('apple-touch-icon.png'),
-      this.destinationPath('app/apple-touch-icon.png')
-    );
+    const _self = this;
```
silvenon: Setting `_self = this` should not be needed, this is the main purpose of arrow functions: you should be able to use `this`, it is inherited from the parent scope.

```js
const _self = this
const fn = () => {
  console.log(this === self) // true
}
```

app/index.js
```diff
@@ -84,146 +42,39 @@ module.exports = class extends Generator {
...
+    const data = {
```
silvenon: A more meaningful variable name, for example:
```suggestion
    const templateData = {
```

app/lib/config.js
```diff
@@ -0,0 +1,93 @@
...
+      checked: true
+    }]
+  }, {
+    type: 'confirm',
+    name: 'includeJQuery',
+    message: 'Would you like to include jQuery?',
+    default: true,
+    when: answers => !answers.features.includes('includeBootstrap')
+  }],
+  dirToCreate: ['app/images', 'app/fonts'],
```
silvenon:
```suggestion
  dirsToCreate: ['app/images', 'app/fonts'],
```

app/index.js
```diff
@@ -84,146 +42,39 @@ module.exports = class extends Generator {
...
+      includeBootstrap: this.includeBootstrap,
+      includeBabel: this.options.babel,
+      testFramework: this.options['test-framework'],
+      includeJQuery: this.includeJQuery,
+      includeModernizr: this.includeModernizr
+    };
+
+    // Render Files
+    config.filesToRender.forEach(file => {
+      helper.copyTpl.call(_self, file.input, file.output, data);
```
silvenon: Nice JavaScript skills!

I was wondering, since there are so few helpers, maybe you could define them in this scope as arrow functions? That way you have them all in one place and you don't have to use `.call()` every time.

app/index.js
```diff
@@ -236,4 +87,4 @@ module.exports = class extends Generator {
       skipInstall: this.options['skip-install']
     });
   }
-};
+};
```
silvenon, line 108: Let's turn on final newlines in your text editor 😄 Which one are you using? Generally it should be enough to install an EditorConfig extension, because we have it configured in `.editorconfig`.

app/lib/helper.js
```diff
@@ -0,0 +1,27 @@
+const mkdirp = require('mkdirp');
+
+function createDirectory(path) {
+  mkdirp(path);
+}
```
silvenon: You can simply do:

```js
const createDirectory = require('mkdirp')
```

but I suggest sticking to the name `mkdirp` because it more accurately signals what it does, i.e. `mkdir -p`, which creates a directory and all its parent directories (if they don't already exist).

silvenon (changes requested):
This is looking great, I love that you extracted repetitive stuff in `config.js`!

app/index.js
UlisesGascon: OMG! So true! Let me refactor this 😄

app/index.js
UlisesGascon: Great! Let me change that 😄

app/index.js
UlisesGascon: 😓 yeah! I am using c9.io and right now is not supported, but I can fix with eslint... let me check that point also in #741

app/lib/helper.js
UlisesGascon: `mkdirp` sounds awesome to me ^^

app/lib/config.js
UlisesGascon: Ok! :-)

app/index.js
UlisesGascon: Sounds good!

UlisesGascon:
@silvenon time to check 😎

UlisesGascon:
Sorry @silvenon I managed wrong some changes between the branches and the last two commits are related to merge conflicts with #739, as now is part of Master Branch. Next Time I will follow the guides xD

silvenon (approved):
LGTM!

Next, I think it's best to add Prettier to unify code style. I'll add it.

silvenon:
Ok, I pushed some changes to master regarding code style. I'm not sure if your editor respects git hooks, if not, ESLint and Prettier errors will now be detected with Travis. 🚀

PR #753, opened 2019-03-01
Update nunjucks.md

jaysunwalter123:
This could do with improvement I'm sure but it's working for me thus far. I'm not at all a js pro, I slowly worked through it, converting from the old recipe as well as I could. Because I don't fully understand what every little bit does my commenting is a little incomplete. Hopefully this will help anyone else that relies on this setup that I personally have used for years.

### DIFF
--- a/docs/recipes/nunjucks.md
+++ b/docs/recipes/nunjucks.md
@@ -60,24 +60,24 @@ Create `app/index.njk`, where you can paste the `<div class="container">` part f
 ### 4. Create a `views` task
 
 ```js
-gulp.task('views', () => {
-  return gulp.src('app/*.njk')
+function views() {
+  return src('app/*.njk')
     .pipe($.nunjucksRender({
       path: 'app'
     }))
-    .pipe(gulp.dest('.tmp'))
-    .pipe(reload({stream: true}));
-});
+    .pipe(dest('.tmp'))
+    .pipe(server.reload({stream: true}));
+};
 ```
 
 This compiles `app/*.njk` files into static `.html` files in the `.tmp` directory.
 
-### 5. Create a 'views:reload' task
+### 5. Create a 'viewsreload' function
 
 ```js
-gulp.task('views:reload', ['views'], () => {
-  reload();
-});
+function viewsreload(){
+  server.reload();
+};
 ```
 
 This triggers Browsersync after `views` task is completed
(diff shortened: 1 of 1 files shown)


docs/recipes/nunjucks.md
```diff
@@ -101,91 +101,69 @@ gulp.task('serve', () => {
...
-+     .pipe(gulp.dest('app/layouts'));
-  });
-```
+  watch([
+-   'app/*.html',
+    'app/images/**/*',
+    '.tmp/fonts/**/*'
+  ]).on('change', server.reload);
 
++ watch('app/**/*.{html,njk}', views);
```
silvenon, line 149: You probably meant to call `viewsreload` here, but instead I suggest removing `viewsreload` function entirely and simply calling `server.reload` directly:

```js
watch('app/**/*.{html,njk}', server.reload);
```

docs/recipes/nunjucks.md
```diff
@@ -101,91 +101,69 @@ gulp.task('serve', () => {
...
+  watch('app/scripts/**/*.js', scripts);
+  watch('app/fonts/**/*', fonts);
+}
+```
 
-### 10. Edit your `serve` task
+### 9. Edit your `serve` task
 
-Edit your `serve` task so that editing `.html` and `.njk` files triggers the `views:reload` task:
+Edit your `serve` task so that editing `.html` and `.njk` files triggers the `viewsreload` function:
```
silvenon, line 158: This description needs to be updated to something more accurate:

> Compile `.njk` files before running the server

But instead I suggest deleting this section and moving the code below to step 6, which contains old `gulpfile.js` code and needs to be updated anyway. You can update its title so the whole 6th step section goes something like this:

> ### 6. Add `views` as a dependency of both `serve` and `build`
>
> ```diff
>  if (isDev) {
> -  serve = series(clean, parallel(styles, scripts, fonts), startAppServer);
> +  serve = series(clean, parallel(views, styles, scripts, fonts), startAppServer);
>  } ...
> ```
>
> ```diff
>  const build = series(
>    parallel(
>     lint,
> -   series(parallel(styles, scripts), html),
> +   series(parallel(views, styles, scripts), html),
>     images,
>     fonts,
>     extras
>    ),
>    measureSize
>  );
> ```

docs/recipes/nunjucks.md
```diff
@@ -101,91 +101,69 @@ gulp.task('serve', () => {
...
-      gulp.watch('app/fonts/**/*', ['fonts']);
-      gulp.watch('bower.json', ['wiredep', 'fonts']);
-    });
-  });
+let serve;
+if (isDev) {
+- serve = series(clean, parallel(styles, scripts, fonts), startAppServer);
++ serve = series(clean, parallel(views, styles, scripts, fonts), startAppServer);
+} ...
+}
```
silvenon, line 166: Best remove this extra `}` to avoid confusion.

silvenon (changes requested):
Thanks Jason! This is looking pretty good, some more updates are necessary, though 😃

UlisesGascon:
Maybe this PR was the same as #766. Should we remix and merge them in a separate branch, @silvenon ? I can handle that if you agree 👍

silvenon:
Since we merged another PR that updated the Nunjucks recipe, I'll close this, but congratulations for the contribution @jaysunwalter123!

PR #758, opened 2019-05-23
Update browserify.md

GallardoCode:
- Fenced code block has console set as language.
- Update script tasks. 
- Babelify config now uses @babel/preset-env preset.
- Updated eslint config.

### DIFF
--- a/docs/recipes/browserify.md
+++ b/docs/recipes/browserify.md
@@ -7,7 +7,7 @@ With this setup you can import modules into your `main.js`. Only code used in `m
 
 ### 1. Install the required plugins
 
-```
+```console
 $ npm install --save-dev browserify babelify vinyl-buffer vinyl-source-stream
 ```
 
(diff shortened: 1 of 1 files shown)


docs/recipes/browserify.md
```diff
@@ -7,7 +7,7 @@ With this setup you can import modules into your `main.js`. Only code used in `m
 
 ### 1. Install the required plugins
 
-```
+```console
```
silvenon: `console` is not a recognized language, AFAIK. Where you thinking of `sh`/`bash`?
```suggestion
```sh
```

docs/recipes/browserify.md
```diff
@@ -24,25 +24,26 @@ $ npm install --save-dev browserify babelify vinyl-buffer vinyl-source-stream
...
+function scripts() {
+-  return src('app/scripts/**/*.js')
 +  const b = browserify({
 +    entries: 'app/scripts/main.js',
-+    transform: babelify,
 +    debug: true
-+  });
++  })
++  .transform(babelify.configure({
++    presets: ["@babel/preset-env"]
```
silvenon: Since generator-webapp comes with `.babelrc` which contains `@babel/preset-env`, is configuring babelify here necessary? Won't it just pick up config from `.babelrc`?

GallardoCode: You are completely right, even if set other babelify options leaving out the preset it will still pick up the config from  `.babelrc`.

silvenon (approved):
Great, thanks! 🚀

Issue #759, opened 2019-06-03
[Road Map] New Release


UlisesGascon:
Hi all!

I opened this issue to keep tracking and an open discussion about the next release features and priorities.

### New features & refactors
- [x] Upgrade outdated dependencies in the generated project **PR:#762**
- [x] Upgrade outdated dependencies in the generator  **PR:#760**
- [ ] Let's improve testing in the generated project
  - [x] Improve the prompt message from Mocha generator #545
  - [x] Allow no test-framework to be used #457
  - [ ] Refactor Jasmine (remove sub-generator)
  - [ ] Refactor Mocha (remove sub-generator)
  - [x] Add jest support #615
  - [ ] Add ava support #521
  - [x] Add Cypress support for e2e testing
- [ ] Let's add a `readme.md` in generated project with relevant info (tasks, etc..) #587
- [ ] Add support to CLI no-interaction install #567 & #458

### Bugs to be fixed
- [ ] Even though Yarn is detected, it creates a npm lockfile #747
- [x] Husky tasks needs to be compatible with NPM and Yarn **PR:#761**
- [x] Update `travis.yml` in order to remove compatibility with Node@6 and add support for Node@12. [Test fall with yarn](https://travis-ci.org/yeoman/generator-webapp/jobs/540656581) **PR:#761**

UlisesGascon:
I will open a PR for the testing part soon, also for dependencies ;-)

UlisesGascon:
@silvenon .. what do you think about this testing workflow? Open to discussion for all of you as well! 😉 
![test_demo](https://user-images.githubusercontent.com/5110813/58906337-ff5a2a80-870b-11e9-9399-82c891c04161.gif)

I will improve the Unit testing with a real testing example (Mocha, Jest, Ava & Jasmine), as well I will cover the example with Cypress for e2e.

_Note: You can ignore the creation files log, as this is only a test for the workflow and not a real implementation yet_

silvenon:
Btw, I updated the name of the milestone to v4.0.0 because so far we have pre-releases, which means that the next stable release will be v4.0.0.

silvenon:
> what do you think about this testing workflow?

I like it!

UlisesGascon:
Awesome! v4.0 then :-)

I will prepare the PR asap 🎉

silvenon:
@UlisesGascon could you explain these two tasks?

> - [ ] Refactor Jasmine (remove sub-generator)
[...]

PR #760, opened 2019-06-03
Upgrade outdated dependencies in the generator

UlisesGascon:
Generator dependencies updated as requested in #759

```
 yeoman-generator         ^3.1.1  →   ^4.0.1 
 @babel/core              ^7.2.2  →   ^7.4.5 
 eslint                  ^5.10.0  →  ^5.16.0 
 eslint-config-prettier   ^3.3.0  →   ^4.3.0 
 eslint-plugin-prettier   ^3.0.0  →   ^3.1.0 
 husky                    ^1.2.1  →   ^2.3.0 
 lint-staged              ^8.1.0  →   ^8.1.7 
 mocha                    ^5.2.0  →   ^6.1.4 
 prettier                ^1.15.3  →  ^1.17.1 
 yeoman-assert            ^3.0.0  →   ^3.1.1 
 yeoman-test              ^1.6.0  →   ^2.0.0 
```

PR #761, opened 2019-06-03
Update ci

UlisesGascon:
As requested in #759:
- Added support for NPM in Husky tasks
- Added support for Node@12 
- Removed for Node@6

### DIFF
--- a/package.json
+++ b/package.json
@@ -53,8 +53,8 @@
   "husky": {
     "hooks": {
       "pre-commit": "lint-staged",
-      "pre-push": "yarn test",
-      "pre-publish": "yarn test"
+      "pre-push": "yarn test || npm test",
+      "pre-publish": "yarn test || npm test"
     }
   },
   "lint-staged": {
--- a/package.json
+++ b/package.json
@@ -53,8 +53,8 @@
   "husky": {
     "hooks": {
       "pre-commit": "lint-staged",
-      "pre-push": "yarn test || npm test",
-      "pre-publish": "yarn test || npm test"
+      "pre-push": "yarn test",
+      "pre-publish": "yarn test"
     }
   },
   "lint-staged": {
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,6 @@
 sudo: false
 language: node_js
 node_js:
+  - v12
   - v10
   - v8
-  - v6
--- a/yarn.lock
+++ b/yarn.lock
@@ -10,33 +10,33 @@
     "@babel/highlight" "^7.0.0"
 
 "@babel/core@^7.2.2":
-  version "7.2.2"
-  resolved "https://registry.yarnpkg.com/@babel/core/-/core-7.2.2.tgz#07adba6dde27bb5ad8d8672f15fde3e08184a687"
-  integrity sha512-59vB0RWt09cAct5EIe58+NzGP4TFSD3Bz//2/ELy3ZeTeKF6VTD1AXlH8BGGbCX0PuobZBsIzO7IAI9PH67eKw==
+  version "7.4.5"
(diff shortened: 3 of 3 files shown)


package.json
```diff
@@ -53,8 +53,8 @@
   "husky": {
     "hooks": {
       "pre-commit": "lint-staged",
-      "pre-push": "yarn test",
-      "pre-publish": "yarn test"
+      "pre-push": "yarn test || npm test",
+      "pre-publish": "yarn test || npm test"
```
silvenon: If someone's making commits and pushes for this project without having Yarn installed, it means that they're maintaining an untracked `package-lock.json` file instead of updating the `yarn.lock` file in our version control, which isn't a good practice. Which package manager a project will use is a decision that everyone needs to stick with, we can't use both.

So unless there is a reason for this that I'm not seeing, I suggest reverting this.

silvenon (changes requested):

package.json
silvenon: Oh, I see what the reason is, I just read #759 again. We have to make Travis use Yarn, that's something we should fix instead. Do you know why this is happening? Maybe something missing from `travis.yml`?

UlisesGascon: As far [I can see...](https://travis-ci.org/yeoman/generator-webapp/jobs/540656581) is a problem related to our upgrade of dependencies.

```
yarn install v1.3.2
[1/5] Validating package.json...
[2/5] Resolving packages...
an-generator@^4.0.1an-generator@^4.0.1e@1.1.3ver@^5.6.0ocha@^6.1.4an-assert@^3.1.1bol-support-x@^1.4.1[3/5] Fetching packages...
error make-dir@3.0.0: The engine "node" is incompatible with this module. Expected version ">=8".
error Found incompatible module
info Visit https://yarnpkg.com/en/docs/cli/install for documentation about this command.
The command "eval yarn " failed 3 times.
The command "yarn" failed and exited with 1 during .
```
One or more modules required node@8, this is why I suggested to remove support for node@6.

silvenon: I totally agree with losing Node v6 👍 I was asking about the 2nd change, where you fall back to `npm test` if `yarn test` fails.

UlisesGascon: Ah @silvenon, I think I got your point now! Well... I needed to add support for Npm as I am not using Yarn in this machine. As we are using Husky (i love it)... I wasn't able to push commits until `yarn test` pass... so I needed to add `npm test` for contributors/maintainers that are not using yarn (like myself) ;-)

silvenon: Is there a specific reason why you can't use it for this project? Imagine that you updated a package to version which has a crucial patch, I will still be stuck with the old version because `yarn.lock` stayed the same.

I used to try sticking with npm myself in Yarn projects, but it doesn't work as soon as you try to manage dependencies. Is not using Yarn a choice or a restriction for you?

silvenon: Because you do manage dependencies in #760, but not `yarn.lock`, which means that as soon as I pull in latest changes and run Yarn, I'll have to make a separate commit for updating the lockfile.

UlisesGascon: I added the missing `yarn.lock` file also to #760 as https://github.com/yeoman/generator-webapp/pull/760/commits/0dc9aebc55d2593ebbe6d4ec97831cfabbf38670. I mostly use only NPM but I don't have problems to use Yarn as well ;-)

silvenon: Do you then agree that we should lose the ` || npm test` part because it gives the false impression that you can use npm too?

UlisesGascon: Yes! Let's stick with Yarn.

UlisesGascon: Added patch in https://github.com/yeoman/generator-webapp/pull/761/commits/e0e625d468bacc31f9a75fa5af5fbe8e2056b64d

silvenon (approved):

Issue #742, opened 2018-12-01
Let's update recipes!


UlisesGascon:
Hi people! 👋 

In the last release... we removed the support for Bower. Many recipes now are very outdated. What do you want to do? Let's discuss it!

- [Current list](https://github.com/yeoman/generator-webapp/tree/master/docs/recipes)  👀

silvenon:
Update one by one, I'm afraid 😄

silvenon:
_Note: @UlisesGascon has updated this list to reflect the ongoing PRs_

### Let's create a checklist:
- [x] Less #767
- [x] Nunjucks #766
- [ ] Handlebars
- [x] Pug #765
- [x] asset revisioning #777
- [ ] AWS S3 deployment
- [ ] browserify
- [ ] CoffeeScript
- [x] ~~Compass~~ · recipe removed in #776 because it's [deprecated](https://github.com/Compass/compass/commit/4de01475c98449c4435359544db443fd3da6e9d3)
- [x] GitHub Pages #778
- [x] HTMLHint #779
- [ ] Liquify
- [ ] Heroku
- [ ] React Router
- [ ] React
- [x] Rsync deploy #775
- [ ] Stylus

jaysunwalter123:
Nunjucks... pleeeeeeeeease 🤓

jaysunwalter123:
Ok, I've updated the gulpfile.js for Nunjucks integration. Where and how can I submit it?

diegoarce:
Is it possible to prioritize the checklist and maybe try to update one per month or something like that. I would love to have the Handlebars recipe.

f97-2308:
Yeah... Pug can't config on current version...  Update it, please... :)))

silvenon:
@jaysunwalter123 by opening a PR to modify [`docs/recipes/nunjucks.md`](https://github.com/yeoman/generator-webapp/blob/master/docs/recipes/nunjucks.md). 🚀

silvenon:
Sorry about not updating recipes, it's not a trivial process, I'll try to update a few soon, starting with Handlebars and Pug.

ly4nhkiet:
please update liquifly, thanks

GlebBigma:
Pug?

silvenon:
@GlebBigma feel free to add feedback in #765.

ilicmarko:
I am really happy that people are using these docs still. Therefore, I will put my time to update most (as much as I can).

[...]

PR #767, opened 2019-10-21
docs(less): update recipe

ilicmarko:
I removed the Bootstrap section as it outdated. Didn't replace it with the new one as I think it is not part of the recipe. Bootstrap will be imported like all other modules.

Read more: yeoman/generator-webapp#742


UlisesGascon (approved):
LGTM! If you agree @silvenon I will merge it 😉 

Thanks a lot for your contribution @ilicmarko 👍

silvenon:
I added some additional fixes. Thanks @ilicmarko!

PR #770, opened 2020-02-22
lint infinite loop

### LINKED ISSUE #769: lint infinite loop.
I have noticed when running the test server the lint goes into an infinite loop, perhaps it is related to this line below.  we watch for a script change, and then lint and the linting process changes the script etc.

https://github.com/yeoman/generator-webapp/blob/c62391b8a58237d9f52f2763497d2e6c049e5833/app/templates/gulpfile.js#L217

erniehs-zz:
Fixes #769.

### DIFF
--- a/app/templates/gulpfile.js
+++ b/app/templates/gulpfile.js
@@ -212,6 +212,7 @@ function startTestServer() {
   });
 
   watch('test/index.html').on('change', server.reload);
+  watch('app/scripts/**/*.js', scripts);
   watch('test/spec/**/*.js', lintTest);
 }
 
--- a/app/templates/gulpfile.js
+++ b/app/templates/gulpfile.js
@@ -211,8 +211,8 @@ function startTestServer() {
     }
   });
 
-  watch('app/scripts/**/*.js', scripts);
-  watch(['test/spec/**/*.js', 'test/index.html'], lintTest);
+  watch('test/index.html').on('change', server.reload);
+  watch('test/spec/**/*.js', lintTest);
 }
 
 function startDistServer() {
--- a/app/templates/gulpfile.js
+++ b/app/templates/gulpfile.js
@@ -89,20 +89,19 @@ async function modernizr() {
 }
 <%_ } -%>
 
-const lintBase = files => {
+const lintBase = (files, options) => {
   return src(files)
-    .pipe($.eslint({ fix: true }))
+    .pipe($.eslint(options))
     .pipe(server.reload({stream: true, once: true}))
     .pipe($.eslint.format())
     .pipe($.if(!server.active, $.eslint.failAfterError()));
 }
 function lint() {
-  return lintBase('app/scripts/**/*.js')
+  return lintBase('app/scripts/**/*.js', { fix: true })
     .pipe(dest('app/scripts'));
 };
 function lintTest() {
-  return lintBase('test/spec/**/*.js')
-    .pipe(dest('test/spec'));
+  return lintBase('test/spec/**/*.js');
 };
 
 function html() {
@@ -213,7 +212,7 @@ function startTestServer() {
   });
 
   watch('app/scripts/**/*.js', scripts);
-  watch(['test/spec/**/*.js', 'test/index.html']).on('change', server.reload);
+  watch(['test/spec/**/*.js', 'test/index.html'], lintTest);
 }
 
 function startDistServer() {
--- a/app/templates/gulpfile.js
+++ b/app/templates/gulpfile.js
@@ -214,7 +214,6 @@ function startTestServer() {
 
   watch('app/scripts/**/*.js', scripts);
   watch(['test/spec/**/*.js', 'test/index.html']).on('change', server.reload);
-  watch('test/spec/**/*.js', lintTest);
 }
 
 function startDistServer() {


silvenon:
Thanks for the PR! While this technically solves the problem of the infinite loop, it also stops linting test files and leaves the task `lintTest` unused.

Instead I propose modifying `lintTest` to stop autofixing by making `lintBase` configurable:

```js
const lintBase = (files, options) => {
  return src(files)
    .pipe($.eslint(options)) // options are passed to ESLint
    // ...
}
function lint() {
  return lintBase('app/scripts/**/*.js', { fix: true }) // configure autofix
    .pipe(dest('app/scripts'));
};
// lintTest stays as-is
```

What do you think?

silvenon (changes requested):
Thanks for the PR! While this technically solves the problem of the infinite loop, it also stops linting test files, which is the intention of `lintTest`, and leaves it unused.

Instead I propose modifying `lintTest` to stop autofixing by making `lintBase` configurable:

```js
const lintBase = (files, options) => {
  return src(files)
    .pipe($.eslint(options)) // options are passed to ESLint
    // ...
}
function lint() {
  return lintBase('app/scripts/**/*.js', { fix: true }) // configure autofix
    .pipe(dest('app/scripts'));
};
// lintTest stays as-is
```

What do you think?

silvenon:
Actually, in `lintTest` we would have to lose the `dest` part to prevent writing the file:

```js
function lintTest() {
  return lintBase('test/spec/**/*.js');
};
```

erniehs-zz:
Perhaps write the lint dest to tmp?

Get Outlook for Android<https://aka.ms/ghei36>

________________________________

silvenon:
It would be misleading to write files that we won't use. Is there a reason why you want to write them somewhere?

erniehs-zz:
Not really I was thinking that they rewrote the files in the editor at times (during infinite loop)  however if loop is gone the fine 🙂

Get Outlook for Android<https://aka.ms/ghei36>

________________________________

silvenon:
Great! Are you willing to add these changes to your PR? And leave the `watch` part.

erniehs-zz:
Hola,

I will have a go implementing these changes

Cheers 🙂

Get Outlook for Android<https://aka.ms/ghei36>

________________________________

app/templates/gulpfile.js
```diff
@@ -213,8 +212,7 @@ function startTestServer() {
   });
 
   watch('app/scripts/**/*.js', scripts);
-  watch(['test/spec/**/*.js', 'test/index.html']).on('change', server.reload);
-  watch('test/spec/**/*.js', lintTest);
+  watch(['test/spec/**/*.js', 'test/index.html'], lintTest);
```
silvenon: Why is this part changed? I suggest returning it to the way it was:

```suggestion
  watch(['test/spec/**/*.js', 'test/index.html']).on('change', server.reload);
  watch('test/spec/**/*.js', lintTest);
```

silvenon (changes requested):

erniehs-zz:
The browser refresh hppens twice as it occurs in lint base.  The server reload part is not needed here it seems.

Ernie

Sent from Mail<https://go.microsoft.com/fwlink/?LinkId=550986> for Windows 10

app/templates/gulpfile.js
erniehs-zz: see below.

silvenon: You're correct that `lintBase` already reloads the server, so the `test/spec/**/*.js` pattern is indeed not needed here, but it doesn't reload when modifying `test/index.html` so I recommend keeping that pattern.

As for your second change, why would changing `test/index.html` need to trigger `lintTest`? Is that on purpose?

silvenon: This is my updated code suggestion:

```suggestion
  watch('test/index.html').on('change', server.reload);
  watch('test/spec/**/*.js'], lintTest);
```

silvenon: Alternatively, you can remove the `server.reload` part from `lintBase`, that doesn't make much sense anyway and it would probably make the reloading behavior more obvious.

I'm fine with either solution.

erniehs-zz: splendid, let me suggest another change in line with this and see where it goes :) we're getting there :D

silvenon: Thanks! ❤️ I hope this is not getting tedious for you.

erniehs-zz: ha ha no probs ;)

app/templates/gulpfile.js
```diff
@@ -212,8 +211,7 @@ function startTestServer() {
     }
   });
 
-  watch('app/scripts/**/*.js', scripts);
```
silvenon, line 215: Please don't remove this.

silvenon (changes requested):
Almost… 🤞😉

silvenon (approved):
Great!

silvenon:
Thank you for spotting this crippling error. Published in [v4.0.0-8](https://github.com/yeoman/generator-webapp/releases/tag/v4.0.0-8). 🚀

PR #771, opened 2020-02-22
add prettier config for auto end of line

erniehs-zz:
to allow windows test run etc.

### DIFF
--- a/package.json
+++ b/package.json
@@ -48,7 +48,8 @@
   "prettier": {
     "semi": true,
     "singleQuote": true,
-    "trailingComma": "none"
+    "trailingComma": "none",
+    "endOfLine": "auto"
   },
   "husky": {
     "hooks": {


package.json
```diff
@@ -48,7 +48,8 @@
   "prettier": {
     "semi": true,
     "singleQuote": true,
-    "trailingComma": "none"
+    "trailingComma": "none",
+    "endOfLine": "auto"
```
silvenon, line 52: Isn't `auto` the default? 🤔 Which problem are you experiencing exactly?

erniehs-zz:
For building on windows boxes it fails the npm test.  Apparently this is not default it seems.

silvenon:
Thanks! 💯 I'll see if I can get my hands on a Windows machine to debug this and possibly report an issue to Prettier 🤷‍♂

PR #776, opened 2020-05-02
Remove compass recipe (deprecated)

antleblanc:
Feel free to tell me if you are agree to remove this recipe because [Compass](https://github.com/Compass/compass/commit/4de01475c98449c4435359544db443fd3da6e9d3
) is no more supported (since Aug, 4 2016).

see:  #742


silvenon:
Agreed, thanks!

PR #777, opened 2020-05-02
docs(recipes): update asset-revisioning

antleblanc:
Update recipe for asset-revisioning.

Replace [gulp-rev-replace](https://github.com/jamesknelson/gulp-rev-replace) _(unmaintained)_ by [gulp-rev-rewrite](https://github.com/TheDancingCode/gulp-rev-rewrite).

see: #742

### DIFF
--- a/docs/recipes/asset-revisioning.md
+++ b/docs/recipes/asset-revisioning.md
@@ -15,24 +15,33 @@ $ npm install --save-dev gulp-rev gulp-rev-replace
 ```
 
 * [gulp-rev](https://github.com/sindresorhus/gulp-rev) appends content hashes
-* [gulp-rev-replace](https://github.com/jamesknelson/gulp-rev-replace) updates references to those files
+* [gulp-rev-rewrite](https://github.com/TheDancingCode/gulp-rev-rewrite) updates references to those files
 
 ### 2. Update the `html` task
 
 Instead of wasting performance reading CSS and JS files into a new stream, we can notice that we already have that stream available in the `html` task, so we can just perform revving there:
 
 ```diff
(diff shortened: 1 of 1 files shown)


docs/recipes/asset-revisioning.md
```diff
@@ -15,24 +15,33 @@ $ npm install --save-dev gulp-rev gulp-rev-replace
...
-  return gulp.src('app/*.html')
+function html() {
+  return src('app/*.html')
     .pipe($.useref({searchPath: ['.tmp', 'app', '.']}))
-    .pipe($.if('*.js', $.uglify()))
-    .pipe($.if('*.css', $.cssnano({safe: true, autoprefixer: false})))
+    .pipe($.if(/\.js$/, $.uglify({compress: {drop_console: true}})))
+    .pipe($.if(/\.css$/, $.postcss([cssnano({safe: true, autoprefixer: false})])))
 +   .pipe($.if('*.js', $.rev()))
 +   .pipe($.if('*.css', $.rev()))
```
silvenon: It seems like it would be more consistent to use regular expressions here as well:
```suggestion
+   .pipe($.if(/\.js$/, $.rev()))
+   .pipe($.if(/\.css$/, $.rev()))
```
Btw, it's really tricky to read diffs of diffs 😆

silvenon (changes requested):
Thanks! I'll assume that this works, so after this small change I'll merge.

docs/recipes/asset-revisioning.md
antleblanc: Thanks! It's now replaced by regular expressions.

silvenon (approved):

Issue #591, opened 2016-11-05
Ditch the diff style in recipes


silvenon:
Even though it's easier to see what to change, it's not copy-paste-friendly.

silvenon:
Might get invalidated by #592.

PR #779, opened 2020-05-06
docs(recipes): update htmlhint

antleblanc:
Update recipe for htmlhint.

see: #742

### DIFF
--- a/docs/recipes/htmlhint.md
+++ b/docs/recipes/htmlhint.md
@@ -17,11 +17,11 @@ $ npm install --save-dev gulp-htmlhint
 Let's create a task in our `gulpfile.js` which runs HTMLHint across all our HTML files and outputs the error in the terminal:
 
 ```js
-gulp.task('htmlhint', () => {
-  return gulp.src('app/**/*.html')
+function htmlhint() {
+  return src('app/**/*.html')
     .pipe($.htmlhint())
     .pipe($.htmlhint.reporter());
-});
+}
 ```
 
 Read [gulp-htmlhint's docs][api] to find out how you can pass options to HTMLHint.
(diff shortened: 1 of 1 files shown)


docs/recipes/htmlhint.md
```diff
@@ -30,25 +30,48 @@ Read [gulp-htmlhint's docs][api] to find out how you can pass options to HTMLHin
 
 HTMLHint should run on serve and build.
 
-```js
-gulp.task('html', ['htmlhint', 'styles', 'scripts'], () => {
+```diff
```
antleblanc, line 33: I moved to diff syntax highlight since I find it more readable.

It's not aligned with what the issue #591 is mentioning but please let me know if you find this acceptable.

Thanks!

silvenon: Here it looks great 👌 I disliked it in recipes with larger changes 😄

docs/recipes/htmlhint.md
```diff
@@ -30,25 +30,48 @@ Read [gulp-htmlhint's docs][api] to find out how you can pass options to HTMLHin
...
- gulp.watch('app/styles/**/*.scss', ['styles']);
- gulp.watch('app/scripts/**/*.js', ['scripts']);
- gulp.watch('app/fonts/**/*', ['fonts']);
- gulp.watch('bower.json', ['wiredep', 'fonts']);
+   watch([
+     'app/*.html',
+     'app/images/**/*',
+     '.tmp/fonts/**/*'
+-  ]).on('change', server.reload);
++  ], {}, htmlhint).on('change', server.reload);
```
silvenon: Is there are reason for passing empty options? They are optional.
```suggestion
+  ], htmlhint).on('change', server.reload);
```

silvenon (changes requested):

docs/recipes/htmlhint.md
antleblanc: No specific reason. I've applied your suggestion.

Thanks!

docs/recipes/htmlhint.md
```diff
@@ -30,25 +30,48 @@ Read [gulp-htmlhint's docs][api] to find out how you can pass options to HTMLHin
...
- gulp.watch('app/styles/**/*.scss', ['styles']);
- gulp.watch('app/scripts/**/*.js', ['scripts']);
- gulp.watch('app/fonts/**/*', ['fonts']);
- gulp.watch('bower.json', ['wiredep', 'fonts']);
+   watch([
+     'app/*.html',
+     'app/images/**/*',
+     '.tmp/fonts/**/*'
+-  ]).on('change', server.reload);
++  ], htmlhint).on('change', server.reload);
```
silvenon: Sorry, I totally missed the more obvious thing 🤦‍♂️ I'd rather create a separate watcher which watches only `app/**/*.html` files and runs `htmlhint` instead of running it here. Because images and fonts can't affect the results of htmlhint, right?

silvenon (changes requested):

docs/recipes/htmlhint.md
silvenon: (By "watcher" I just mean adding another `gulp.watch`.)

antleblanc: I totally agree! 👍
It's now fixed.

silvenon (approved):
Great!

